PR #10, opened 2023-08-22
Reduce incorrectly indexed contracts

### LINKED ISSUE #6: perf: heuristically reject bad contracts emitting events
**Problem**

The watch tower listens for events, that are *not* filtered by a specific contract address. Options:

1. Filter by specific address: This makes the system more brittle, as contracts are redeployed with new features, the watch tower would be dependent on specific changes for the new deployment address.
2. Listen to events, blockchain-wide: This is the current method deployed, but introduces additional overhead by asserting that the contracts implement the required interface.

**Solution**

At the core of the watch tower is the regular checking for orders to be placed via the `getTradeableOrderWithSignature` function. If we were to check only for this function selector (`bytes4`), the number of false check passes may still be high. We can add an additional check for the `cabinet(address,bytes32)` function to reduce the number of false positives and create a more robust heuristic.

mfw78:
This PR:

1. Introduces a `getCode` check to assert that at least the `cabinet` and `getTradeableOrderWithSignature` functions are implemented.
2. Provides basic tests around the `getCode` functions.
3. In doing (2) above, configures a `jest` test harness for unit testing of the watch tower.
4. Includes the *full* `forge` output in the ABI, as `deployedBytecode` is used in the test harnesses.

**Related issues**

- Closes #6


actions/utils.ts
```diff
@@ -328,3 +329,35 @@ export function sendSlack(message: string): boolean {
...
+
+  const composableCow = ComposableCoW__factory.createInterface();
+  for (const signature of signatures) {
+    const sighash = composableCow.getSighash(signature);
+    if (!code.includes(sighash.slice(2))) {
+      return false;
+    }
+  }
+
+  return true;
```
alfetopito, line 362: Just a fun little refactor suggestion, to be functional - in the functional programming sense:

```suggestion
  return signatures.every((signature) => {
    const sighash = composableCow.getSighash(signature)
    
    return code.includes(sighash.slice(2))
  })
```

actions/tsconfig.json
```diff
@@ -9,6 +9,6 @@
     "target": "es2020",
     "resolveJsonModule": true
   },
-  "exclude": ["**/*.spec.ts"],
+  // "exclude": ["**/*.spec.ts"],
```
alfetopito, line 12: Why commenting it out rather than removing?

actions/utils.ts
```diff
@@ -328,3 +329,35 @@ export function sendSlack(message: string): boolean {
...
+ * a good enough heuristic to filter out most of the contracts that do not implement the interface.
+ * 
+ * @dev The selectors are:
+ * - `cabinet(address,bytes32)`: `1c7662c8`
+ * - `getTradeableOrderWithSignature(address,(address,bytes32,bytes),bytes,bytes32[])`: `26e0a196`
+ * 
+ * @param code the contract's deployed bytecode as a hex string
+ * @returns A boolean indicating if the contract likely implements the interface
+ */
+export function compatCheck(code: string): boolean {
```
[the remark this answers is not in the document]
mfw78: Nice suggestion on the name. Opted for `isCompatible`.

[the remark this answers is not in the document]
mfw78: Nice suggestion on the name. Opted for `isCompatible`.

actions/utils.ts
mfw78: Slick :+1:

alfetopito (approved):

actions/tsconfig.json
anxolin: Its good to add I believe, just avoid compiling tests which is not needed

https://stackoverflow.com/questions/36475736/exclude-with-pattern-in-tsconfig-json

actions/utils.ts
```diff
@@ -328,3 +335,27 @@ export function sendSlack(message: string): boolean {
...
+ * a good enough heuristic to filter out most of the contracts that do not implement the interface.
+ * 
+ * @dev The selectors are:
+ * - `cabinet(address,bytes32)`: `1c7662c8`
+ * - `getTradeableOrderWithSignature(address,(address,bytes32,bytes),bytes,bytes32[])`: `26e0a196`
+ * 
+ * @param code the contract's deployed bytecode as a hex string
+ * @returns A boolean indicating if the contract likely implements the interface
+ */
+export function isCompatible(code: string): boolean {
```
anxolin, line 354: nit: I would prefer to give a more specific name to this util. `isComposableCoWCompatible`

uber-nit: would it better to be more explicit here with the var name, sth like `composableCowBytecode`?

actions/addContract.ts
```diff
@@ -41,6 +42,12 @@ const _addContract: ActionFn = async (context: Context, event: Event) => {
   // Process the logs
   let hasErrors = false;
   for (const log of transactionEvent.logs) {
+    // Do not process logs that are not from a `ComposableCoW`-compatible contract
+    // This is a *normal* case, if the contract is not `ComposableCoW`-compatible
+    // then we do not need to do anything, and therefore don't flag as an error.
+    if (!isCompatible(await provider.getCode(log.address))) {
```
anxolin, line 48: If Composable CoW is known, why not checking the address. 
Do we care about other instances of the contracts?

anxolin: The check would be faster, and wouldn't give you false positives, right?

actions/tsconfig.json
mfw78: I may not understand the `jest` foo here, but from my understanding, `jest` doesn't work with typescript, and as the tests are written in typescript, these need to be compiled (otherwise, there would be the need for babel or something to transpile? I left this with the simplest I knew which is just compile everything, including the tests, and then run `jest` on this.

actions/addContract.ts
mfw78: Yes, to both. Ultimately it comes down to the question of if we want to run a generic watch tower, or not. Given that others may end up piggy-backing on our infrastructure (and this leads to more cows), I think it's a fair thing to support. An example here is Swapr and their Stackly product, that was using our old watch Web3 Actions (which we weren't aware of).

actions/model.ts
```diff
@@ -49,12 +49,12 @@ export type ConditionalOrder = {
   proof: Proof | null;
   // a map of discrete order hashes to their status
   orders: Map<OrderUid, OrderStatus>;
-  // the address to poll for orders
-  composableCow: string;
+  // the address to poll for orders (may, or **may not** be `ComposableCoW`)
+  address: string;
```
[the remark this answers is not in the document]
mfw78: Tracking in #12

actions/utils.ts
mfw78: Tracked in #13

actions/utils.spec.ts
```diff
@@ -0,0 +1,32 @@
+import * as composableCow from './artifacts/ComposableCoW.json'
+import * as extensibleFallbackHandler from './artifacts/ExtensibleFallbackHandler.json'
+import { isCompatible } from './utils'
+
+describe('test supports composable cow interface from bytecode', () => {
+    it('should pass', () => {
+        expect(isCompatible(composableCow.deployedBytecode.object)).toBe(true)
```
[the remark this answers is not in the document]
mfw78: Tracked in #13

PR #15, opened 2023-08-23
fix: nethermind

### LINKED ISSUE #11: dep: nethermind returns custom errors in manner not expected by ethers
**Problem**

Gnosis Chain is often used as a testing ground for low value-at-risk trades in a "testing" environment. However, Gnosis Chain is heavily biased towards Nethermind as an execution client, and as such, inherits bugs there. Notably, https://github.com/NethermindEth/nethermind/issues/6024 refers to Nethermind not correctly returning revert codes (as may be expected).

**Solution**

There are two options:

1. Investigate use of `eth_estimateGas` to see if this would return the correct `revert` value.
2. Derive some intermediate step for handling of Gnosis Chain until nodes are migrated to a version of Nethermind that fixes this fundamental issue.

**CC**: @cowprotocol/backend @cowprotocol/frontend

mfw78:
This PR:

1. Provides extended RPC error handling around the Nethermind case (as is most common for Gnosis Chain).

Related issues:

- Closes #11


actions/checkForAndPlaceOrder.ts
```diff
@@ -415,15 +415,58 @@ function _handleGetTradableOrderCall(
...
+    // error selector from `error.error.error.data` and use that to determine the error.
+    let parsedError = "";
+    if (error.data === "0x") {
+      // Verify that `error.error.error.data` is defined and is a string, otherwise it
+      // could be a different error
+      // TODO: Is there a better way to do this?
+      if (
+        error.error &&
+        error.error.error &&
+        typeof error.error.error.data === "string"
```
anxolin, line 441: LOL 🤣
`error.error.error`
`error.success.error`
`error.success.error.error.success.data.error`

actions/checkForAndPlaceOrder.ts
```diff
@@ -415,15 +415,58 @@ function _handleGetTradableOrderCall(
...
+      ) {
+        // 9 skips the `Reverted ` prefix, and 19 is the end of the error selector
+        parsedError = String(error.error.error.data).slice(9, 19);
+      }
+    } else {
+      parsedError = error.errorName; // If it's not nethermind, we can use the error name
+    }
```
anxolin, line 448: should we the hack under a private function so we don't need see the dark magic that retrieves the error?

actions/checkForAndPlaceOrder.ts
```diff
@@ -415,15 +415,58 @@ function _handleGetTradableOrderCall(
...
+
+    switch (parsedError) {
       case "OrderNotValid":
+      case "0xc8fc2725":
```
anxolin, line 452: What is this hex? can you use a const so it doesn't look like a magic value?

actions/checkForAndPlaceOrder.ts
```diff
@@ -415,15 +415,58 @@ function _handleGetTradableOrderCall(
...
         // The conditional order has not expired, or been cancelled, but the order is not valid
         // For example, with TWAPs, this may be after `span` seconds have passed in the epoch.
+
+        // As the `OrderNotValid` is parameterized, we can retrieve the reason. This is an abi encoded string,
+        // so we need to decode it.
+        // TODO: use the `OrderNotValid` reason?
+        const reason = ethers.utils.defaultAbiCoder.decode(
+          ["string"],
+          String(error.error.error.data).slice(19)
```
anxolin, line 461: I feel the `error.error.error.data`, which seems prone to error, is being processed in more than one branch. In one of the branches, you had an if in case `error.error.error.data` was not defined, but not in this one.

Maybe, before the switch we can just process the error and extract the relevant info, and then we enter the switch with clean data

anxolin (commented):
I can sense in this PR Nethermind is not your favorite node.

actions/checkForAndPlaceOrder.ts
mfw78: Magic 💫
It corresponds to the custom error selectors, analogous to the way that function selectors are calculated. `consts` were added with some docs.

actions/checkForAndPlaceOrder.ts
mfw78: This has been refined so that their magic is only handled in one branch, with heavy validation of the assumptions.

actions/checkForAndPlaceOrder.ts
mfw78: Let's not start :laughing:

mfw78:
> I can sense in this PR Nethermind is not your favorite node.

🦦🦦

actions/utils.spec.ts
```diff
@@ -33,4 +33,114 @@ describe('test against concrete examples', () => {
     it('should fail with no selectors', () => {
         expect(isComposableCowCompatible('0xdeadbeefdeadbeef')).toBe(false)
     })
-})
\ No newline at end of file
+})
+
+describe('handle nethermind reversion (simple)', () => {
+
+    const SINGLE_ORDER_NOT_AUTHED_ERROR: any = {
```
alfetopito, line 40: Nitpick: declare the consts outside the test fn, unless it's something that needs the local context
Also, why the type `any`?

actions/utils.ts
```diff
@@ -359,3 +365,103 @@ export function isComposableCowCompatible(code: string): boolean {
...
+ * @returns an empty parsed error if assumptions don't hold, otherwise the selector and message if applicable.
+ */
+const rawErrorDecode = (abi: string): ParsedError  => {
+  if (abi.length == 10) {
+    return { errorNameOrSelector: abi }
+  } else {
+    try {
+      const selector = String(abi).substring(0, 10);
+      console.log('selector', selector)
+      console.log('message', abi.slice(10))
```
alfetopito, line 386: Debug statements?

actions/utils.ts
```diff
@@ -359,3 +365,103 @@ export function isComposableCowCompatible(code: string): boolean {
...
+  message?: string;
+}
+
+/**
+ * Given a raw ABI-encoded custom error returned from a revert, extract the selector and optionally a message.
+ * @param abi of the custom error, which may or may not be parameterised.
+ * @returns an empty parsed error if assumptions don't hold, otherwise the selector and message if applicable.
+ */
+const rawErrorDecode = (abi: string): ParsedError  => {
+  if (abi.length == 10) {
```
alfetopito, line 380: Triple the fun:

```suggestion
  if (abi.length === 10) {
```

actions/utils.ts
```diff
@@ -359,3 +365,103 @@ export function isComposableCowCompatible(code: string): boolean {
...
+    return { errorNameOrSelector: abi }
+  } else {
+    try {
+      const selector = String(abi).substring(0, 10);
```
alfetopito, line 384: `abi` is a string already, I don't think you need to wrap it in `String`

Also, just a note on the method, I've seem mainly `slice` used for this sort of thing, while I don't think I ever saw `substring`.

Comparison: https://stackoverflow.com/a/31910656/1272513

Not that it matters

actions/utils.ts
mfw78: Nice catch, thanks.

actions/utils.ts
mfw78: Yes, it's really ABI then. When passing back a custom error as a reversion message in solidity, it returns the custom error selector, which is formatted exactly the same way as when making a function call (formatting the ABI).

So for example, when calling `deposit` in the `WETH9` contract, the actual function you call is just `deposit()`, there is no parameter as the amount being passed in is in `msg.value`. So, in this case, the ABI is actually only `0xd0e30db0`.

Same case for custom errors that don't have any parameters. :+1:

actions/utils.ts
mfw78: Nice, thanks for the tips :+1:

actions/utils.ts
```diff
@@ -359,3 +365,112 @@ export function isComposableCowCompatible(code: string): boolean {
...
+  } else {
+    // This is a Nethermind node, as `data` *must* be equal to `0x`, but we know we always revert with an
+    // message, so - we have to go digging ⛏️🙄
+    //
+    // Verify our assumption that `error.error.error.data` is defined and is a string.
+    // TODO: Is there a better way to do this?
+    if (
+      error.error &&
+      error.error.error &&
+      typeof error.error.error.data === "string"
```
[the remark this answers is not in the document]
mfw78: Thanks, makes perfect sense that there's a more idiomatic way to do this.

actions/artifacts/Multicall3.json
```diff
@@ -0,0 +1,440 @@
+[
+    {
+        "inputs": [
```
anxolin, line 3: I don't think you need the whole ABI, right?

only aggregate3value?

image

actions/utils.ts
```diff
@@ -359,3 +365,105 @@ export function isComposableCowCompatible(code: string): boolean {
...
+  // custom error. Attempt to decode and return.
+  if (data !== "0x") {
+    console.log(data);
+    return rawErrorDecode(data)
+  } else {
+    // This is a Nethermind node, as `data` *must* be equal to `0x`, but we know we always revert with an
+    // message, so - we have to go digging ⛏️🙄
+    //
+    // Verify our assumption that `error.error.error.data` is defined and is a string.
+    // TODO: Is there a better way to do this?
```
alfetopito, line 446: Does this TODO still apply?

actions/checkForAndPlaceOrder.ts
```diff
@@ -29,6 +34,7 @@ import { GPv2Order } from "./types/ComposableCoW";
 import { validateOrder } from "./handlers";
 
 const GPV2SETTLEMENT = "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
+const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
```
anxolin, line 37: Is it the same address in all networks?

actions/utils.ts
```diff
@@ -359,3 +365,105 @@ export function isComposableCowCompatible(code: string): boolean {
...
+  // - It's not a parameterised custom error - no message
+  // - We can return early
+  if (errorName) {
+    return { errorNameOrSelector: errorName };
+  }
+
+  // If error.data is not zero-bytes, then it's not a Nethermind node, assume it's a string parameterised
+  // custom error. Attempt to decode and return.
+  if (data !== "0x") {
+    console.log(data);
```
anxolin, line 439: pls, remember to delete

actions/utils.ts
```diff
@@ -359,3 +365,105 @@ export function isComposableCowCompatible(code: string): boolean {
...
+    return rawErrorDecode(data)
+  } else {
+    // This is a Nethermind node, as `data` *must* be equal to `0x`, but we know we always revert with an
+    // message, so - we have to go digging ⛏️🙄
+    //
+    // Verify our assumption that `error.error.error.data` is defined and is a string.
+    // TODO: Is there a better way to do this?
+    if (typeof error?.error?.error?.data === "string") {
```
anxolin, line 447: nit: maybe save first `error?.error?.error?.data` into const, so you don't need to repeat in next line

anxolin (approved):
Nice! great you added some tests. 

only nitpicks. This logic should eventually make its way to the SDK, if the polling is done there... 
Also reason why you should keep the ABI small

actions/checkForAndPlaceOrder.ts
mfw78: Yep, it's the same across all networks. Their [repository refers](https://github.com/mds1/multicall).

actions/utils.ts
mfw78: 🤦‍♂️ Done.

actions/artifacts/Multicall3.json
mfw78: Trimmed.

actions/utils.ts
mfw78: Not at all, consider it gone!

PR #16, opened 2023-08-24
Use SDK to poll and create smart order and handle result

anxolin:
# Poll logic
Incorporate all the poll logic from the SDK

## Why the logic is being moved to the SDK?
It is great to delegate to the SDK because:
- Watch Tower doesn't need to know what type of order is indexing. 
   - See this PR for more details https://github.com/cowprotocol/cow-sdk/pull/149
- The polling now will check on basic things like if the order is authorised (so we can remove that part of code from this project)
- The polling will also check on specifics of concrete orders: i.e TWAP could check if the order is expired so we don't try to index again

## Auth Check in the SDK + Don't try again logic
The PR incorporate some handling of the poll result. Like when we don't need to try again to index an order.

Additionally, now by delegating to the SDK, we check out of the box if the order has been authenticated, and prints a nice message:

image

The watch tower will understand that this order doesn't need to be re-attempted in the future. 

In order to test this, you can create an order and cancel it. Or you could index this block number:
```
BLOCK_NUMBER=29707125 
NETWORK=100
```

## In case of success
This version will still successfully post the order to the API:
image

The only difference is that now, the responsibility of who is returning the discrete order details is in the SDK. 

## Network is not a String anymore
This PR changes the network from a `String` to a `SupportedChainId`. Since we use the SDK, it is nice we use its types. 

Sorry for the noise on the chain. 

One place where I didn't change it, is in the `Registry` that is persisted. I haven't update there yet to avoid making an incompatible chain in this PR. I prefer to avoid migrating the model for this.

To reproduce add this to your env
```
NETWORK=100
BLOCK_NUMBER=29707218
```

## Utils refactor
The `utils.ts` was too big, i just broke it done into a few files.

## Not included 

* I have to do a typescript cast because the type we persist in the model don't match what we actually store. We should change the model, but for that we need to add migrations. This is completely out of scope of this PR: https://github.com/cowprotocol/tenderly-watch-tower/issues/18
* Follow up PR will handle other results from the polling, like


actions/poll.ts
```diff
@@ -0,0 +1,26 @@
...
+const ordersFactory = new ConditionalOrderFactory(
+  DEFAULT_CONDITIONAL_ORDER_REGSTRY
+);
+
+export async function pollConditionalOrder(
+  params: PollingParams
+): Promise<PollResult | undefined> {
+  const { owner, chainId, conditionalOrderParams, provider } = params;
+
+  const order = ordersFactory.fromParams(params.conditionalOrderParams);
```
anxolin, line 19: now we just pass blindly the params, and the factory will create the `ConditionalOrder` for us

After that we can use the SDK methods from the order, including of course the polling.

anxolin:
Documenting an issue im having:

If you add this env:
```
NETWORK=100
BLOCK_NUMBER=29490646
```

And run with `yarn start:actions`

You will reproduce this:

This awaited promised were returning the 3 strings we need:
image

Now they return undefined, its like the params is now a string and not an object
image

Reference https://github.com/cowprotocol/tenderly-watch-tower/blob/3be7215b063dfa52091e96d78fb440e4c87fee30/actions/checkForAndPlaceOrder.ts#L170

actions/checkForAndPlaceOrder.ts
```diff
@@ -86,29 +103,49 @@ const _checkForAndPlaceOrder: ActionFn = async (
         chainContext.provider
       );
 
-      const { deleteConditionalOrder, error } = await _processConditionalOrder(
+      const pollResult = await _processConditionalOrder(
```
alfetopito, line 106: I find this variable name misleading.

When truthy it means it's an error, so you only care for this in case it errors.

What about calling it `error` instead?
Or something that makes it clearer

actions/checkForAndPlaceOrder.ts
```diff
@@ -86,29 +103,49 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
+        // TODO: Handle the other errors :) --> Store them in the registry and ignore blocks until the moment is right
+        //  TRY_ON_BLOCK
+        //  TRY_AT_EPOCH
+      }
 
-      if (deleteConditionalOrder) {
-        ordersPendingDelete.push(conditionalOrder);
+      // Log the result
+      const resultDescription =
+        pollResult !== undefined
```
alfetopito, line 130: This check is done on top already.

```suggestion
        error
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -119,91 +156,84 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
+    // const [handler, salt, staticInput] = await (() => {
+    //   const [handler, salt, staticInput ] = conditionalOrder.params;
+    //   return Promise.all([handler, salt, staticInput]);
+    // })();
+    // console.log("TODO: Why now this parameters seem broken????? ", {
+    //   handler,
+    //   salt,
+    //   staticInput,
+    //   params: conditionalOrder.params,
+    // });
```
alfetopito, line 198: Are we keeping this?

actions/test/run_local.ts
```diff
@@ -18,11 +19,12 @@ const main = async () => {
...
 
-  const testRuntime = await _getRunTime(network);
+  const chainId = toChainId(network);
+  const testRuntime = await _getRunTime(chainId);
 
   // Get provider
-  const provider = await getProvider(testRuntime.context, network);
-  const { chainId } = await provider.getNetwork();
+  const provider = await getProvider(testRuntime.context, chainId);
+  await provider.getNetwork();
```
alfetopito, line 27: Not storing the result?

actions/utils/context.ts
```diff
@@ -0,0 +1,212 @@
...
+      return false;
+    });
+}
+
+var consoleOriginal = {
+  log: console.log,
+  error: console.error,
+  warn: console.warn,
+  debug: console.debug,
+};
```
alfetopito, line 212: Why?

actions/utils/misc.ts
```diff
@@ -0,0 +1,53 @@
+import { OrderStatus } from "../model";
+import {
+  ALL_SUPPORTED_CHAIN_IDS,
+  SupportedChainId,
+} from "@cowprotocol/cow-sdk";
+
+type LocalChainId = 31337;
+const LOCAL_CHAIN_ID = 31337;
```
alfetopito, line 8: Does this work?

```suggestion
const LOCAL_CHAIN_ID = 31337;
type LocalChainId = typeof LOCAL_CHAIN_ID;
```

alfetopito (approved):

actions/checkForAndPlaceOrder.ts
anxolin: You are right!

actions/checkForAndPlaceOrder.ts
anxolin: i was, but now i deleted...

actions/utils/context.ts
anxolin: good question, cause before i was doing some custom logging, but now i can delete. 

Deleted!

actions/test/run_local.ts
anxolin: Nup, we don't need. It's just a quick check to see if its working t. This is for the local run...

I believe it was already there, but i will delete.

actions/checkForAndPlaceOrder.ts
```diff
@@ -119,91 +167,80 @@ const _checkForAndPlaceOrder: ActionFn = async (
     }
   }
 
+  // Delete owners with no orders
+  for (const [owner, conditionalOrders] of Array.from(ownerOrders.entries())) {
+    if (conditionalOrders.size === 0) {
+      ownerOrders.delete(owner);
+    }
+  }
+
```
mfw78, line 176: This is quite valid, and most likely exactly what we want to do. Another thing is that having the keys for the `ownerOrders` (ie. previous safes that have traded), allows us to possibly use this as some cache for valid safes that are likely to have upgraded their fallback handler etc. If you can think of a use case for this, it may be worthwhile keeping around the keys, otherwise we can bin them.

actions/checkForAndPlaceOrder.ts
```diff
@@ -119,91 +167,80 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
-    // Get GPv2 Order
-    const tradeableOrderResult = await _getTradeableOrderWithSignature(
-      owner,
-      network,
-      conditionalOrder,
-      contract,
-      multicall
+      salt,
+    };
+    let pollResult = await pollConditionalOrder(
```
mfw78, line 220: Maybe want some way to pass through offChainInput / proofs.

mfw78 (approved):
An epic review. A few nits, and comments specifically dealing with merkle roots of orders.

actions/checkForAndPlaceOrder.ts
```diff
@@ -57,24 +62,41 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
+    chainId,
     context
   );
   const { ownerOrders } = registry;
 
   // enumerate all the owners
   let hasErrors = false;
-  console.log(`[checkForAndPlaceOrder] New Block ${blockEvent.blockNumber}`);
+  let ownerCount = 0;
+  let orderCount = 0;
```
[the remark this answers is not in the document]
anxolin: Tbh, this was just cause theres a lot of output, and I didn't want to print the params multiple time, so I just added a prefix that makes it a bit easier to see all the logs related to the same order.

I used a counter to not add a long text in the log... 

This is not the SDK, just watch tower, i don't think the people will appreciate a JSDOc in this local vars. Since they are counters, i think with having that in the name would suffice. 

Maybe i missed the point, let me know otherwise

actions/model.ts
```diff
@@ -2,10 +2,11 @@ import Slack = require("node-slack");
 
 import { Context, Storage } from "@tenderly/actions";
 import { Transaction as SentryTransaction } from "@sentry/node";
-import { BytesLike, ethers } from "ethers";
+import { BytesLike, ethers, providers } from "ethers";
 
 import { apiUrl, getProvider } from "./utils";
 import type { IConditionalOrder } from "./types/ComposableCoW";
```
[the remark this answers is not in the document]
anxolin: I think this asks for another PR, i will make an issue. This PR has too many MISC changes: https://github.com/cowprotocol/tenderly-watch-tower/issues/19

actions/utils.ts
```diff
@@ -169,15 +174,16 @@ function getAuthHeader({ user, password }: { user: string; password: string }) {
   return "Basic " + Buffer.from(`${user}:${password}`).toString("base64");
 }
 
-export function apiUrl(network: string): string {
-  switch (network) {
-    case "1":
+// TODO: If we use the Ordebook  API a lot of code will be deleted. Out of the scope of this PR (a lot has to be cleaned)
```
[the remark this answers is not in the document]
anxolin: this was deleted

actions/checkForAndPlaceOrder.ts
```diff
@@ -455,8 +450,8 @@ function _handleGetTradableOrderCall(
         // TODO: Make use of `message` returned by parseCustomError ?
 
         return {
-          result: ValidationResult.FailedButIsExpected,
-          deleteConditionalOrder: false,
+          result: PollResultCode.TRY_NEXT_BLOCK,
+          reason: "OrderNotValid",
```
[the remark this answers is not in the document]
anxolin: But this is what we do, right? 

if the value is `OrderNotValid` (or the selector of the same error), we return as the reason `OrderNotValid`

Or what do you mean?

actions/checkForAndPlaceOrder.ts
anxolin: Good point

actions/checkForAndPlaceOrder.ts
anxolin: I will comment out for now, and we decide later

I think one use case, is just to avoid doing the verification. I just added this to not keep adding owners and big loops over time. I would hope, once the order is executed or expired, we don't need to keep things around. 

Leaving it for now

PR #21, opened 2023-09-05
Misc: Persist polling result, poll latest block, versioning support

anxolin:
# Misc changes
This PR makes some small improvements, in principle not very related, but to keep things simpler, i prefer to do in a single PR

## Log the Conditional Order using the SDK
The conditional orders have a `toString` implementation that would show in plain English the order details. This is super handy to understand any issues with the Polling Logic that come after:
image

I added the `pollResult` to the conditional order, with the latest poll result.

## Add a basic versioning to the conditional order model
Although this PR is backward compatible with the model, a future PR might not be. 

image

This is why I want to introduce in this PR some versioning on the conditional order model.

The key is the revived function, it will receive the current version that needs to be parsed. It's expected that the function will parse the registry and convert it to the latest version, effectively applying a model migration. 

In this PR, because there's no breaking change, we can just ignore the version number that is passed as a parameter. 

## Allow to index once for the latest block
One feature of this project is that you can run it once for a specific block .
Another one, is we can place the storage from tenderly in the `STORAGE` env (to reproduce Tenderly state)

If you experience an issue in Tenderly indexing and you want to run locally, its convenient to copy the storage, and then run it on the latest block.

This PR allows to provide `BLOCK_NUMBER=latest` in the env to do this.

# Not included
This PR is preparation for the next one, that will handle the `pollResult` errors, to ignore the indexing of some blocks

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,7 +9,9 @@ NETWORK=5 # Network to test. 1: Mainnet, 5: Goerli, 100: xDai
 # Optionally, specify the block number to process locally
 #   - useful for debugging or re-apply an old block to make sure the orders are created)
 #   - if not specified, it will run in WATCH mode, subscribing to all new blocks
+#   - Accepts either "latest" for the latest block, or the block number
 #BLOCK_NUMBER=9500767
+#BLOCK_NUMBER=latest
 BLOCK_NUMBER=
 
 # Slack
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -108,7 +108,7 @@ const _checkForAndPlaceOrder: ActionFn = async (
         chainContext.provider
       );
 
-      const pollError = await _processConditionalOrder(
+      const pollResult = await _processConditionalOrder(
         owner,
         chainId,
         conditionalOrder,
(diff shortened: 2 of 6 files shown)


actions/model.ts
```diff
@@ -188,3 +228,12 @@ export function replacer(_key: any, value: any) {
     return value;
   }
 }
+function parseConditionalOrders(
+  serializedConditionalOrders: string | undefined,
+  _version: number | undefined
```
alfetopito, line 233: Why the version if it's not used?

actions/checkForAndPlaceOrder.ts
```diff
@@ -118,38 +118,36 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
-        //  TRY_ON_BLOCK
-        //  TRY_AT_EPOCH
+      // Don't try again the same order, in case thats the poll result
+      if (pollResult.result === PollResultCode.DONT_TRY_AGAIN) {
+        ordersPendingDelete.push(conditionalOrder);
       }
 
+      // Save poll result
+      conditionalOrder.pollResult = {
+        lastExecution: new Date(),
```
mfw78, line 129: We may want to be careful with the use of `Date()` here as multiple orders that are updated in the same block will all have different execution times. This may be inconsistent with expected logic.

Suggest that for `lastExecution` we reference the block.timestamp.

mfw78 (approved):
Minor nit RE: timestamp consistency

actions/checkForAndPlaceOrder.ts
anxolin: That's a great suggestion. Actually, i can add also the block number

.env.example
```diff
@@ -9,7 +9,9 @@ NETWORK=5 # Network to test. 1: Mainnet, 5: Goerli, 100: xDai
 # Optionally, specify the block number to process locally
 #   - useful for debugging or re-apply an old block to make sure the orders are created)
 #   - if not specified, it will run in WATCH mode, subscribing to all new blocks
+#   - Accepts either "latest" for the latest block, or the block number
 #BLOCK_NUMBER=9500767
+#BLOCK_NUMBER=latest
 BLOCK_NUMBER=
```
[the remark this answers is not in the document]
anxolin: If empty, it will do the normal watch-mode, and index all new blocks. This is only meaningful for running it locally

actions/checkForAndPlaceOrder.ts
```diff
@@ -254,9 +254,10 @@ async function _processConditionalOrder(
 
     const orderToSubmit: Order = {
       ...order,
-      kind: kindToString(order.kind),
-      sellTokenBalance: balanceToString(order.sellTokenBalance),
-      buyTokenBalance: balanceToString(order.buyTokenBalance),
+      kind: kindToString(order.kind.toString()),
+      sellTokenBalance: balanceToString(order.sellTokenBalance.toString()),
+      buyTokenBalance: balanceToString(order.buyTokenBalance.toString()),
```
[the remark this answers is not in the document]
anxolin: yes, i think so :)

PR #22, opened 2023-09-05
Handle some API errors

anxolin:
This PR aims to reduce the noise on errors happening during the polling of conditional orders.

A good default handling is to return `TRY_NEXT_BLOCK` for the errors:
  * InsufficientBalance,
  * InsufficientAllowance,
  * InsufficientFee

These errors are handled since we don't want to consider this as an error, and retrying in next block might give a different result.

In the future, we can do some better error handling by delegating to the SDK (the SDK can use the API, and also delegate to child orders)

This PR turns green DEV environment:
image

### DIFF
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -97,10 +97,7 @@ const _checkForAndPlaceOrder: ActionFn = async (
     blockNumber
   );
 
-  console.log(
-    "[checkForAndPlaceOrder] Number of orders: ",
-    Array.from(registry.ownerOrders.values()).flatMap((s) => s).length
-  );
+  console.log("[checkForAndPlaceOrder] Number of orders: ", registry.numOrders);
 
   for (const [owner, conditionalOrders] of ownerOrders.entries()) {
     ownerCounter++;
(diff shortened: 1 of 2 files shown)


actions/checkForAndPlaceOrder.ts
```diff
@@ -85,6 +97,11 @@ const _checkForAndPlaceOrder: ActionFn = async (
     blockNumber
   );
 
+  console.log(
+    "[checkForAndPlaceOrder] Number of orders: ",
+    Array.from(registry.ownerOrders.values()).flatMap((s) => s).length
+  );
```
anxolin: Log the number of orders. Happy to remove it, for now its helpful to see clearly the number of orders at the beginning and end of the execution

actions/checkForAndPlaceOrder.ts
```diff
@@ -184,7 +198,7 @@ const _checkForAndPlaceOrder: ActionFn = async (
   //   registry.stringifyOrders()
   // );
 
-  Array.from(registry.ownerOrders.values()).flatMap((s) => s);
+  console.log("[checkForAndPlaceOrder] Remaining orders: ", registry.numOrders);
```
anxolin, line 201: just something that was handy during my testings

mfw78 (approved):

PR #23, opened 2023-09-05
Add loggly

anxolin:
Test to log into a different tool: Loggly

They index the logs, which is nice, because they will identify for example the parameters from the order:
image

this PR adds also some tags to facilitate the filtering. 
We can filter by the tenderly operation, for example `addContract`, but also by chain `chain_100`:
image

We can filter by the values of specific  fields, and they show the different values they have for the field, for example, here we can actually see how many errors we get from the API and which type:
image

We can easily filter by owner for example too:
image

## Log only if error
For now, I'm in the 30 days trial period, but when is over, we will need to pay for the usage

If we log over 1GB a day is super expensive. This is easy to supus given we are verbose and we index block for 3 chains. 

This PR includes a flag we can enable in the future to:
- Accumulate logs in a buffer
- Only print those logs if we get at a later point an error. This will serve as an "context for that error"

### DIFF
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -9,7 +9,7 @@ import {
 import axios from "axios";
 
 import { ethers } from "ethers";
-import { BytesLike, Logger, hexlify } from "ethers/lib/utils";
+import { BytesLike, Logger } from "ethers/lib/utils";
 
 import {
   ComposableCoW,
--- a/actions/utils/context.ts
+++ b/actions/utils/context.ts
@@ -218,7 +218,7 @@ async function _initLogging(
 ) {
   const logglyToken = await context.secrets.get("LOGGLY_TOKEN").catch(() => "");
   if (logglyToken) {
-    initLogging(logglyToken, [transactionName, `chain:${chainId}`]);
+    initLogging(logglyToken, [transactionName, `chain_${chainId}`]);
   } else {
     console.warn("LOGGLY_TOKEN is not set, logging to console only");
   }
--- a/actions/utils/context.ts
+++ b/actions/utils/context.ts
@@ -21,7 +21,9 @@ export async function initContext(
   chainId: SupportedChainId,
   context: Context
 ): Promise<ExecutionContext> {
-  initLogging();
+  // Init Logging
+  await _initLogging(transactionName, chainId, context);
+
   // Init registry
   const registry = await Registry.load(context, chainId.toString());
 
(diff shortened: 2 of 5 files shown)


mfw78 (approved):
This is cool!

actions/checkForAndPlaceOrder.ts
```diff
@@ -9,7 +9,7 @@ import {
 import axios from "axios";
 
 import { ethers } from "ethers";
-import { BytesLike, Logger, hexlify } from "ethers/lib/utils";
+import { BytesLike, Logger } from "ethers/lib/utils";
```
alfetopito, line 12: Isn't the eslint rule to not allow unused imports on?

actions/utils/context.ts
```diff
@@ -203,3 +207,19 @@ function handlePromiseErrors<T>(
...
+  transactionName: string,
+  chainId: SupportedChainId,
+  context: Context
+) {
+  const logglyToken = await context.secrets.get("LOGGLY_TOKEN").catch(() => "");
+  if (logglyToken) {
+    initLogging(logglyToken, [transactionName, `chain_${chainId}`]);
+  } else {
+    console.warn("LOGGLY_TOKEN is not set, logging to console only");
+  }
```
alfetopito, line 224: Since this fails silently, how can we tell when loggly is turned on in production?

actions/utils/logging.ts
```diff
@@ -0,0 +1,99 @@
...
+    tags,
+    json: true,
+  });
+
+  if (logger !== undefined) {
+    // We had a previous logger, remove it and add the new one
+    winston.remove(logger);
+    winston.add(newLogger);
+    logger = newLogger;
+    return;
```
alfetopito, line 24: Why returning when it already exists?
Is the rest of the setup not needed?

alfetopito (commented):
Couple of questions

actions/utils/context.ts
anxolin: We can see it in Tenderly log, or in local. 
I don't care too much, we can see it by logs not being pushed to Loggly

actions/utils/logging.ts
anxolin: I just do this to change the tags. The rest uses the `winston.log` which use the new logger.

anxolin: One execution could init the logging multiple times, this is why i did it

alfetopito (approved):

PR #31, opened 2023-09-20
Optimise stored state

### LINKED ISSUE #30: perf: remove owners with no orders
### Background
When observing the performance of the tenderly watch tower in production, as time went on, it was observed that the storage has become less and less reliable (as noted by logs in loggly showing saving of registry fails occasionally). This has often been worked around by running multiple watch towers (sub-optimal solution).

When observing the production state for mainnet, the registry is 120kb in size. There are two reasons for this bloat:

1. Removal of conditional orders from the watch tower was not straight-forward without the now-implemented TWAP-specific logic.
2. A decision was not yet taken on whether or not to remove owners from the registry with no orders (can retain the fact that they are _likely_ composable-cow enabled.

### Details

Storage space should be limited to an absolute bare minimum given the seeming high cost of this storage API. Therefore, it's desired to:

1. Correctly prune defunct conditional orders (TWAP-specific logic): 120kb --> 17kb. (Implemented)
2. Prune owners with no conditional orders: 17kb -> 12kb. (This issue).

### Acceptance criteria

- [x] Owners with no conditional orders are pruned from the registry.

mfw78:
# Description
The watch tower has in production had timeout issues with storage. Therefore, storage should be assumed to be super expensive to call / retrieve, so we should minimise the amount of state that we are storing.

# Changes
If an owner no longer has any conditional orders being monitored - remove them from the registry.

- [x] Remove owners with no orders being monitored.

## How to test
Using PR #29 run a local store rebuild and observe that the end state returned does not contain any owners with no orders.

## Related Issues

Closes #30

### DIFF
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -208,15 +208,13 @@ const _checkForAndPlaceOrder: ActionFn = async (
     }
   }
 
-  // TODO: Decide if we want to delete owners with no orders
-  //    - One one hand, we don't want to keep growing our registry forever
-  //    - On the other hand, it might be handy to know this owner has been verified before
-  // // Delete owners with no orders
-  // for (const [owner, conditionalOrders] of Array.from(ownerOrders.entries())) {
-  //   if (conditionalOrders.size === 0) {
-  //     ownerOrders.delete(owner);
-  //   }
-  // }
+  // It may be handy in other versions of the watch tower implemented in other languages
+  // (ie. not for Tenderly) to not delete owners, so we can keep track of them.
+  for (const [owner, conditionalOrders] of Array.from(ownerOrders.entries())) {
+    if (conditionalOrders.size === 0) {
+      ownerOrders.delete(owner);
+    }
+  }
 
   // Update the registry
   hasErrors ||= await !writeRegistry();


actions/checkForAndPlaceOrder.ts
```diff
@@ -208,15 +208,13 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
-  //    - One one hand, we don't want to keep growing our registry forever
-  //    - On the other hand, it might be handy to know this owner has been verified before
-  // // Delete owners with no orders
-  // for (const [owner, conditionalOrders] of Array.from(ownerOrders.entries())) {
-  //   if (conditionalOrders.size === 0) {
-  //     ownerOrders.delete(owner);
-  //   }
-  // }
+  // It may be handy in other versions of the watch tower implemented in other languages
+  // (ie. not for Tenderly) to not delete owners, so we can keep track of them.
```
alfetopito, line 212: Why would we want to keep track of former owners?
Is that for analytics purposes or something else?

alfetopito (approved):

anxolin (approved):

actions/checkForAndPlaceOrder.ts
mfw78: It can be for analytics as an example, but also for other optimisations. At the moment, when we add a contract, we retrieve it's code with `getCode` and check for the presence of some solidity functions within. If we leveraged an event watcher for `ChangedFallbackHandler`, and watched for changes on the `owners` we would have a way of optimising this code so that when existing owners add more conditional orders, we don't have to do this `getCode` again and save on some RPC calls.

Would require some benchmarking to see the effects on RPC call utilisation though.

PR #32, opened 2023-09-20
Allow to index single tx and improve logs

anxolin:
# Description
After reviewing the logs, and the cases pointed out by @mfw78 on unindexed order I created this PR to:
- Improve logging
- Make it easier to re-process an order for a given TX

## Reduced error messages
Handled errors are not errors :) 
So for example, if the order is already orderbook and we just need to wait, is not an error. So i reduced to WARN those messages.

This hopefully will let us filter the errors in the day and reduce the noise.

See https://github.com/cowprotocol/tenderly-watch-tower/pull/32/files#diff-8e2dc90f4a6ebdf01f8603707f37f7794972fdf1827102bfce89497804f72837R185

## Allow to re-process a TX
Now you can easily debug cases of TX that create orders by adding the TX hash in the `.env`

This is way faster for debugging cases, since processing a whole block is very slow (process 100-200tx!)

https://github.com/cowprotocol/tenderly-watch-tower/pull/32/files#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8cR19

## Reduced messages

Don't write messages that don't add a lot of value:
https://github.com/cowprotocol/tenderly-watch-tower/pull/32/files#diff-5d3bc67362ae1598162466e069b336740bfdb4f706d88e8057b0e85132210438R72

## Let addContract log the TX
When debugging in the logs is convenient to search by blockNumber or hash.

`checkForAndPlaceOrder` already logs the TX, but not the `addContract`. 
It was hard to find a specific order. So this PR adds the TX and block into the logs

https://github.com/cowprotocol/tenderly-watch-tower/pull/32/files#diff-5d3bc67362ae1598162466e069b336740bfdb4f706d88e8057b0e85132210438R147

## Force SEQUENTIAL execution
Its a requirement for this project right now, so i made it explicit:
https://github.com/cowprotocol/tenderly-watch-tower/pull/32/files#diff-e3185fd2787bf4c6056698f3a82e0403ada878137d9d82346d3d1d12c2640556R11

### DIFF
--- a/actions/test/run_local.ts
+++ b/actions/test/run_local.ts
@@ -215,14 +215,11 @@ async function processBlock(
     }
   }
 
-  const pollingBlock = overrides?.blockWatchBlockNumber
-    ? await provider.getBlock(blockNumber)
-    : block;
-
   hasErrors ||= !(await _pollAndPost({
-    block: pollingBlock,
+    block,
     chainId,
     testRuntime,
+    blockWatchBlockNumber: overrides?.blockWatchBlockNumber,
   }));
 
   if (hasErrors) {
@@ -325,16 +322,18 @@ async function _pollAndPost({
   block,
   chainId,
   testRuntime,
+  blockWatchBlockNumber,
 }: {
   block: ethers.providers.Block;
   chainId: number;
   testRuntime: TestRuntime;
+  blockWatchBlockNumber?: number;
 }) {
   const blockNumber = block.number;
 
   // Block watcher for creating new orders
   const testBlockEvent = new TestBlockEvent();
-  testBlockEvent.blockNumber = blockNumber;
+  testBlockEvent.blockNumber = blockWatchBlockNumber ?? blockNumber;
   testBlockEvent.blockDifficulty = block.difficulty?.toString();
   testBlockEvent.blockHash = block.hash;
   testBlockEvent.network = chainId.toString();
--- a/actions/test/run_local.ts
+++ b/actions/test/run_local.ts
@@ -204,13 +204,13 @@ async function processBlock(
       overrides?.txList?.includes(transaction.hash) ?? true;
 
     if (shouldProcessTx) {
-      hasErrors ||= await !_processTx(
+      hasErrors ||= !(await _processTx(
         provider,
         block,
         chainId,
         testRuntime,
         transaction
-      );
+      ));
     }
   }
 
@@ -235,10 +235,7 @@ async function processTx(
   chainId: number,
   testRuntime: TestRuntime
 ) {
-  // Execute once, for a specific tx
-  console.log(
-    `[run_local] Processing ONCE for a specific transaction: ${tx}...`
-  );
+  let hasErrors = false;
   const transaction = await provider.getTransaction(tx);
   if (!transaction.blockNumber) {
     throw new Error(`The transaction ${tx} is not mined yet (no blockNumber)`);
@@ -248,8 +245,18 @@ async function processTx(
     throw new Error(`[run_local] Transaction ${tx} not found`);
   }
 
-  await _processTx(provider, block, chainId, testRuntime, transaction);
-  await _pollAndPost({ block, chainId, testRuntime });
+  hasErrors ||= !(await _processTx(
+    provider,
+    block,
+    chainId,
+    testRuntime,
+    transaction
+  ));
+  hasErrors ||= !(await _pollAndPost({ block, chainId, testRuntime }));
+
+  if (hasErrors) {
+    throw new Error("[run_local] Errors found in processing TX");
+  }
 }
 
 async function _processTx(
--- a/actions/test/run_local.ts
+++ b/actions/test/run_local.ts
@@ -8,7 +8,11 @@ import { addContract } from "../addContract";
 import { ethers } from "ethers";
 import assert = require("assert");
 import { toChainId, getProvider } from "../utils";
-import { ProcessBlockOverrides, ReplayPlan, getOrdersStorageKey } from "../model";
+import {
+  ProcessBlockOverrides,
+  ReplayPlan,
+  getOrdersStorageKey,
+} from "../model";
 import { exit } from "process";
 import { SupportedChainId } from "@cowprotocol/cow-sdk";
 import { ComposableCoW__factory } from "../types/factories/ComposableCoW__factory";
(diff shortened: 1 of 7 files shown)


actions/test/run_local.ts
```diff
@@ -25,9 +25,24 @@ const main = async () => {
...
   const blockNumberEnv = process.env.BLOCK_NUMBER;
-  if (blockNumberEnv) {
+  if (txEnv) {
+    // Execute once, for a specific tx
+    console.log(
+      `[run_local] Processing ONCE for a specific transaction: ${txEnv}...`
+    );
+
+    await processTx(provider, txEnv, chainId, testRuntime).catch(() => {
+      exit(101);
```
alfetopito, line 41: Why exit code `101`? Is this a standard or is this defined anywhere? (Sorry, not aware)

actions/test/run_local.ts
```diff
@@ -25,9 +25,24 @@ const main = async () => {
...
 
-  // Run one of the 2 Execution modes (single block, or watch mode)
+  // Run one of the 3 Execution modes:
+  //  - Single tx: Process just one tx
+  //  - Single block: Process just one block
+  //  - Watch mode: Watch for new blocks, and process them
+  const txEnv = process.env.TX;
   const blockNumberEnv = process.env.BLOCK_NUMBER;
-  if (blockNumberEnv) {
+  if (txEnv) {
```
alfetopito, line 34: Maybe mention in the docs/readme the order of precedence for the env vars?

actions/test/run_local.ts
```diff
@@ -72,58 +87,124 @@ async function processBlock(
...
-          );
-          return false;
-        });
-      console.log(
-        `[run_local] Result of "addContract" action for TX ${hash}: ${_formatResult(
-          result
-        )}`
-      );
-    }
+    hasErrors ||= !(await !_processTx(
```
alfetopito, line 90: Why the double negation?
Or the `!` before the promise means something different?

actions/test/run_local.ts
```diff
@@ -72,58 +87,124 @@ async function processBlock(
...
+      chainId,
+      testRuntime,
+      transaction
+    ));
+  }
+
+  hasErrors ||= !(await _pollAndPost({ block, chainId, testRuntime }));
+
+  if (hasErrors) {
+    throw new Error("[run_local] Errors found in processing block");
```
alfetopito, line 102: Does it make sense to also log the block number where this failed for extra clarity?

```suggestion
    throw new Error(`[run_local] Errors found in processing block: ${block}`);
```

actions/test/run_local.ts
```diff
@@ -72,58 +87,124 @@ async function processBlock(
...
+  chainId: number,
+  testRuntime: TestRuntime
+) {
+  let hasErrors = false;
+  const transaction = await provider.getTransaction(tx);
+  if (!transaction.blockNumber) {
+    throw new Error(`The transaction ${tx} is not mined yet (no blockNumber)`);
+  }
+  const block = await provider.getBlock(transaction.blockNumber);
+  if (!transaction) {
```
alfetopito, line 118: If the transaction can be falsy, line 114 can fail. Better add the `?.` operator.
Or move the `transaction` check up

```suggestion
  const transaction = await provider.getTransaction(tx);
  if (!transaction?.blockNumber) {
    throw new Error(`The transaction ${tx} is not mined yet (no blockNumber)`);
  }
  const block = await provider.getBlock(transaction?.blockNumber);
  if (!transaction) {
```

actions/test/run_local.ts
```diff
@@ -72,58 +87,124 @@ async function processBlock(
...
+    provider,
+    block,
+    chainId,
+    testRuntime,
+    transaction
+  ));
+  hasErrors ||= !(await _pollAndPost({ block, chainId, testRuntime }));
+
+  if (hasErrors) {
+    throw new Error("[run_local] Errors found in processing TX");
```
alfetopito, line 132: Add tx id for extra info

```suggestion
    throw new Error(`[run_local] Errors found in processing TX: ${tx}`);
```

actions/test/run_local.ts
```diff
@@ -72,58 +87,124 @@ async function processBlock(
...
+      transactionHash: transaction.hash,
+    };
+
+    // run action
+    const result = await testRuntime
+      .execute(addContract, testTransactionEvent)
+      .then(() => true)
+      .catch((e) => {
+        console.error(
+          `[run_local] Error running "addContract" action for TX:`,
```
alfetopito, line 180: Also add the hash here:

```suggestion
          `[run_local] Error running "addContract" action for TX: ${hash}`,
```

alfetopito (commented):
Couple of suggestions

actions/test/run_local.ts
anxolin: haha, no. Just to make it different than 100 I used for the other. I will just define consts for them

actions/test/run_local.ts
anxolin: done

actions/test/run_local.ts
anxolin: great catch!

actions/test/run_local.ts
anxolin: It does!

actions/test/run_local.ts
anxolin: But why would it be falsy? it will throw if there's no TX, if there is, it will give me the receipt. Right?

anxolin: successful promise will give the Tx Response...

actions/test/run_local.ts
```diff
@@ -72,57 +88,123 @@ async function processBlock(
...
+) {
+  let hasErrors = false;
+  const transaction = await provider.getTransaction(tx);
+  if (!transaction.blockNumber) {
+    throw new Error(`The transaction ${tx} is not mined yet (no blockNumber)`);
   }
+  const block = await provider.getBlock(transaction.blockNumber);
+  if (!transaction) {
+    throw new Error(`[run_local] Transaction ${tx} not found`);
+  }
```
alfetopito, line 121: If `getTransaction` throws when not found, `transaction` will never be falsy.
Or will it?

anxolin: According to Typescript, is possible. This is why I assert `!transaction.blockNumber`

- The TX response cant be falsy, either throws or it returns the receipt
- the blockNumber can, see image:
 
image

anxolin:
Consolidating, let me know if there's any further comments.

This PR also fixes some issues in main branch. Currently broken (fixed when rebasing)

To take into account in #34

PR #33, opened 2023-09-20
Handle race condition

anxolin:
# Description
This PR tries to avoid this situaton we observed.

This order created in this TX https://etherscan.io/tx/0x06f595b3e60257ef7bfb168e53e52a5527993c22fbe5fe2278936902da1a2fd8  was not being monitored by the Watch Tower because, according to it, it EXPIRED 🕣
image

Re-running the Watch Tower locally for the same TX gave a totally different result
image

Question is, how is it possible the first execution said the order was EXPIRED, but a later one said it was not? 🤔

The answer is the CABINET. This is where TWAP orders save the `mining time`. 

...but if the cabinet is persisted in the same TX where we create the order (create with context)... How is it possible that we registered the order, but there's nothing in the cabinet, but later there's something in the cabinet? 🤔

One possible explanation is:
* Watch Tower uses different RPC provider as we do
* They are the ones that notifies us when they see a new event of creation of a composable cow order
* We registered using their information
* When we checked later in our own RPC provider, we saw an empty cabinet. It could be that that TX was not yet in our RPC node, because we are lagging behind!

## Changes in this PR
This PR will avoid doing anything drastic (deleting the order from their watch list), if the TX that registered the order is not in their blockchain state.

Initially I was going to enforce some number of confirmation before I delete orders, but I think this should be equally reliably and it would cause less noise (i don't want to re-attempt checking on an expired order, if indeed its expired)


alfetopito:
Can you point this to https://github.com/cowprotocol/tenderly-watch-tower/pull/32 to isolate what's changed only in this PR?

anxolin:
> Can you point this to https://github.com/cowprotocol/tenderly-watch-tower/pull/32 to isolate what's changed only in this PR?

Sorry, done!

actions/checkForAndPlaceOrder.ts
```diff
@@ -161,7 +161,18 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
+        // Check if the TX exists. This avoids one race condition where the order has been registered but our RPC don't see it yet.
+        // It can happen if:
+        //    - Tenderly RPC node is ahead of our RPC node
+        //    - There's a reorg between the registration and the polling
+        const existTransaction = await chainContext.provider
+          .getTransaction(conditionalOrder.tx)
+          .then(() => true)
+          .catch(() => false);
+
+        if (existTransaction) {
```
alfetopito, line 173:
```suggestion
        const transactionExists = await chainContext.provider
          .getTransaction(conditionalOrder.tx)
          .then(() => true)
          .catch(() => false);

        if (transactionExists) {
```

actions/checkForAndPlaceOrder.ts
```diff
@@ -161,7 +161,18 @@ const _checkForAndPlaceOrder: ActionFn = async (
...
       if (pollResult.result === PollResultCode.DONT_TRY_AGAIN) {
-        ordersPendingDelete.push(conditionalOrder);
+        // Check if the TX exists. This avoids one race condition where the order has been registered but our RPC don't see it yet.
+        // It can happen if:
+        //    - Tenderly RPC node is ahead of our RPC node
+        //    - There's a reorg between the registration and the polling
+        const existTransaction = await chainContext.provider
+          .getTransaction(conditionalOrder.tx)
+          .then(() => true)
+          .catch(() => false);
```
alfetopito, line 171: I'm wondering: Is it possible by any chance that it returns undefined when the tx doesn't exist rather than throwing an exception?

alfetopito (approved):
A question and a nitpick

actions/checkForAndPlaceOrder.ts
anxolin: I don't want it to throw an exception, i want to:

- existTransaction be true if I know for sure it exists
- be false other wise (if it doesn't exist, or even if Im unsure, like in an unhandled exception)

I use this to avoid deletion of the order. I want to make sure we know what we are doing

mfw78:
Will not port these changes to the `standalone` branch as it's execution does step on any database key read/write, and there is no RPC race condition as it triggers and executes based on the same RPC provider (same "view of truth").

actions/checkForAndPlaceOrder.ts
alfetopito: That's my question.
Can `getTransaction` return something that's falsy without throwing?
The result from `getTransaction` is being ignored and it returns `true` regardless of the value.

PR #39, opened 2023-09-21
feat: commander

### LINKED ISSUE #35: chore: use commander for environment / cli configuration

mfw78:
# Description
Introduces commander to deal with the various program run modes and unifying configuration. 

# Changes

- [x] Add commander
- [x] Template the location for the run logic
- [x] Move types to `types/generated`, relocate `model` and define some interfaces.

## How to test

To review:

Run the CLI and observe the help. Subsequent programs should exit throwing "Not Implemented".

```bash
yarn ts-node ./src/index.ts --help
```

## Related Issues

Closes #35


src/watchtower.ts
```diff
@@ -0,0 +1,76 @@
...
+      "ComposableCoW contract address"
+    )
+    .requiredOption(
+      "--deployment-block <deploymentBlock...>",
+      "Block number at which the contracts were deployed"
+    )
+    .requiredOption(
+      "--page-size <pageSize>",
+      "Number of events to fetch per page",
+      "5000"
```
alfetopito, line 26: Given it has a default, does it need to be a required option?

src/watchtower.ts
```diff
@@ -0,0 +1,76 @@
...
+    )
+    .option("--publish", "Publish orders to the OrderBook API", true)
+    .action((options: RunOptions) => {
+      // Need to assert that the RPCs and deployment blocks are the same length
+      const { rpc, deploymentBlock } = options;
+
+      if (rpc.length !== deploymentBlock.length) {
+        throw new Error("RPC and deployment blocks must be the same length");
+      }
```
alfetopito, line 35: That requirement is not mentioned in the option descriptions.
Also, now I realized the multiple RPCs and deployment blocks are meant to be 1 per network, but that's not clear from the options.

What about instead the format is changed a bit to make it more clear and avoid a mistake I would certainly make to pass the blocknumber for a different network given that the order has to be followed in two different parameter groups.

Something like: `--chainInfo: 1983712:https://rpc.bla.bla` where the first part is the block number and the second is the rpc endoint.

src/watchtower.ts
anxolin: But is this the number of events or blocks?

src/modes/replay.ts
```diff
@@ -0,0 +1,9 @@
+import { ReplayBlockOptions, ReplayTxOptions } from "../types";
+
+export async function replayBlock(_options: ReplayBlockOptions) {
```
anxolin, line 3: nit: just a suggestion, but since there's various commands, I would make a dir commands and one file per command. 
Here replay defines 2 commands, and there might be commands that are not in the "modes" dir, so it feels easier to mantain if theres 1:1 mapping between command and file

package.json
```diff
@@ -7,15 +7,16 @@
...
   "scripts": {
-    "build": "rm -rf dist && rm -rf types && yarn typechain && tsc",
+    "build": "rm -rf dist && rm -rf src/types/generated && yarn typechain && tsc",
     "fmt": "prettier --write \"src/**/*.ts\"",
     "lint": "eslint \"src/**/*.ts\" && prettier --check \"src/**/*.ts\"",
     "lint:fix": "eslint --fix \"src/**/*.ts\" && prettier --write \"src/**/*.ts\"",
     "test": "yarn build && jest ./dist",
-    "typechain": "typechain --target ethers-v5 --out-dir src/types/ \"abi/*.json\"",
+    "typechain": "typechain --target ethers-v5 --out-dir src/types/generated/ \"abi/*.json\"",
     "prepare": "husky install && yarn typechain"
```
anxolin, line 16: can you add a script to run the cli easily? 

Also, we should probably add a bin section https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin

src/types/index.ts
```diff
@@ -0,0 +1,22 @@
+export interface WatchtowerOptions {
+  contract: string;
+  publish: boolean;
```
anxolin, line 3: what is the publish option? maybe good to add some JS docs

src/types/index.ts
```diff
@@ -0,0 +1,22 @@
+export interface WatchtowerOptions {
```
anxolin, line 1: nit: consider using this `index.ts` to reexports all the types from model and any other file in `/types` dir. 

Also, consider moving all these types to `./cli.ts` or `./api.ts` or `./watchTower.ts`

src/watchtower.ts
mfw78: Made an `option`, and indeed it is related to blocks. Fixed.

src/watchtower.ts
mfw78: Added help text before the options in relation to the RPC / deployment block orders.

Will consider the `--chainInfo` with future PRs that are likely to allow monolithic (watch all chains), and also "run an instance per chain".

src/watchtower.ts
```diff
@@ -0,0 +1,76 @@
...
+  program
+    .command("run")
+    .description("Run the watchtower")
+    .requiredOption("--rpc <rpc...>", "Chain RPC endpoints to monitor")
+    .requiredOption(
+      "--contract <contractAddress>",
+      "ComposableCoW contract address"
+    )
+    .requiredOption(
+      "--deployment-block <deploymentBlock...>",
```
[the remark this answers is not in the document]
mfw78: This is scoped for upcoming PRs looking at variadic environment variables. I'm very hesitant to use config files as these place additional considerations on dockerising the environment. `env` works well all-around :smiley:

src/watchtower.ts
```diff
@@ -0,0 +1,76 @@
+import { program } from "@commander-js/extra-typings";
+import { ReplayBlockOptions, ReplayTxOptions, RunOptions } from "./types";
+import { replayBlock, replayTx, run } from "./modes";
+
+async function main() {
+  program
+    .name("watchtower")
+    .description("Monitoring composable cows on the blockchain 🐮")
```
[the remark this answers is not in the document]
mfw78: Done, excluding documentation reference link.

src/modes/replay.ts
mfw78: Yep, I think this is where things will naturally converge. Great idea. Tracking in #60.

package.json
mfw78: Tracking in #60

src/types/index.ts
mfw78: Tracking in #60

src/types/index.ts
mfw78: Tracking in #60.

PR #44, opened 2023-09-21
feat: leveldb

### LINKED ISSUE #40: feat: use leveldb for persistence

mfw78:
# Description
Given that we are already using a NoSQL database and we need some storage, I dropped in `leveldb` as the NoSQL solution. This allows us to eliminate all external dependencies, making it friendlier for spinning up in bandwidth constrained environments (ie. hackathons), and reduces complexity with deployment (no requirement for external SQL servers).

# Changes

- [x] Removed all references to Tenderly context / storage.
- [x] Added `level` packages and `ts.
- [x] Created `db` singleton for holding the database.
- [x] Removed all secrets as these are now defined by `commander`.
- [x] Refactored init to remove `await` that were dependent on `secret` lookups previously.

## How to test

No testing for this PR, besides `eslint` and `prettier` passing, excluding to-be-deprecated actions. This PR is mostly for "readability".

## Related Issues

Closes #40


src/types/model.ts
```diff
@@ -124,18 +129,37 @@ export class Registry {
...
+    const lastNotifiedError = await storage
+      .getDB()
+      .get(getNetworkStorageKey(LAST_NOTIFIED_ERROR_STORAGE_KEY, network))
+      .then((isoDate: string | number | Date) =>
+        isoDate ? new Date(isoDate) : null
+      )
+      .catch(() => null);
+
+    const lastProcessedBlock = await storage
+      .getDB()
```
alfetopito, line 150: What about getting the db instance only once and re-using it in all the following calls?

src/utils/db.ts
```diff
@@ -0,0 +1,48 @@
...
+  protected db: DBLevel;
+
+  private static _instance: DBService = new DBService();
+
+  protected constructor() {
+    if (DBService._instance) {
+      throw new Error(
+        "Error: Instantiation failed: Use DBService.getInstance() instead of new."
+      );
+    }
```
alfetopito, line 15: What about making it a singleton?
If the instance already exists just return it rather than throwing

src/utils/db.ts
```diff
@@ -0,0 +1,48 @@
...
+        valueEncoding: "json",
+        createIfMissing: true,
+        errorIfExists: false,
+      });
+    } else {
+      this.db = new Level<string, string>("./database", {
+        valueEncoding: "json",
+        createIfMissing: true,
+        errorIfExists: false,
+      });
```
alfetopito, line 29: Nitpick: options are identical. Create a const for it an use it on both

src/utils/db.ts
```diff
@@ -0,0 +1,48 @@
+import { Level } from "level";
+
+export type DBLevel = Level<string, string>;
+
+export default class DBService {
```
anxolin, line 5: nit: This name is a bit generic (but it depeds on Level db).

It would be great if you can use the Repository or DAO pattern to abstract the persistance. This way, this implementation will use Level DB, but others could use anything else.

This could have a repository or DAO layer, and then you create a `ContextRepository` interface, with a `ContextRepositoryLevel` or `ContextRepositoryImpl` (since we don't have more)

src/utils/db.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  protected constructor() {
+    if (DBService._instance) {
+      throw new Error(
+        "Error: Instantiation failed: Use DBService.getInstance() instead of new."
+      );
+    }
+    DBService._instance = this;
+
+    if (process.env.NODE_IS_TEST === "true") {
```
anxolin, line 18: why not using NODE_ENV

also, i don't think the this class should be aware of env vars, it should receive all params in the constructor

src/utils/db.ts
```diff
@@ -0,0 +1,48 @@
...
+  public async open() {
+    await this.db.open();
+  }
+
+  public async close() {
+    await this.db.close();
+  }
+
+  public getDB() {
+    return this.db;
```
anxolin, line 46: If you use the repository pattern, you would not expose this, but would provide for the service layer the basic CRUD operations you need

anxolin (approved):
Great PR!

anxolin (approved):
Great PR! 

Adding some comments, no need to address for this PR, but it would be great to consider for the top of the waterfall

anxolin (approved):
Great PR!

Could you ignore the db in git?
image

src/utils/db.ts
```diff
@@ -0,0 +1,48 @@
...
+        "Error: Instantiation failed: Use DBService.getInstance() instead of new."
+      );
+    }
+    DBService._instance = this;
+
+    if (process.env.NODE_IS_TEST === "true") {
+      this.db = new Level<string, string>("./database_test", {
+        valueEncoding: "json",
+        createIfMissing: true,
+        errorIfExists: false,
```
anxolin, line 22: all this logic feels duplicated. The only difference i see between the 2 is the db name

why not doing:
```ts
const dbName = process.env.NODE_IS_TEST ? 'database_test' : 'database'
his.db = new Level<string, string>(dbName, {
        valueEncoding: "json",
        createIfMissing: true,
        errorIfExists: false,
      });
```

src/types/model.ts
mfw78: Done

src/utils/db.ts
mfw78: Refactored to now receive database location from the constructor.

src/utils/db.ts
mfw78: Paths are passed to the constructor, ready for stitching in with env aware logic, however will defer this to #60.

src/utils/db.ts
mfw78: Now respects paths passed via the constructor.

PR #45, opened 2023-09-21
Schedule not balance orders + fix issue with no due orders

anxolin:
# Description

This PR is an attempt to handle cases when the orders don't have enough balance or allowance better

## Reduce the checks
Because is a run-time error, we will actually re-attempt in next block. 

This will create an error in the logs every 15seconds!
Further more, we will be consuming RPC calls, and punish the API trying to post orders that are unbacked!

This PR will handle this case, and schedule next poll only after 10 minutes (this should reduce errors and calls by `50x` and still give the user a chance to add the balance and allowance to get their order back on track)

This handling is less drastic than the alternative that would be to return a `DONT_TRY_AGAIN`.

## Fix issue with not due orders
While reviewing the logs after these cleanups, i realised the condition was flipped! so we were ignoring orders that should be checked. 

This PR also fixes this. 

## Improve logs
This case will be nicely logged

image

## Misc
Additionally, now we use an enum for the errors instead of hardcoded strings

image

## Handle case: Don't show as ERROR
- Before this PR they show as unhandled ERROR, since we actually try to post them to the API and we get a run-time error
- After this PR it will log them as WARN

# Context

https://cowservices.slack.com/archives/C05RMJB7ZFA/p1695308323236179

### DIFF
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -119,10 +119,12 @@ const _checkForAndPlaceOrder: ActionFn = async (
       // Check if the order is due (by epoch)
       if (
         lastResult?.result === PollResultCode.TRY_AT_EPOCH &&
-        lastResult.epoch < blockTimestamp
+        blockTimestamp < lastResult.epoch
       ) {
         console.log(
-          `${logPrefix} Skipping conditional. Reason: Not due yet (TRY_AT_EPOCH=${lastResult.epoch}). ${logOrderDetails}`,
+          `${logPrefix} Skipping conditional. Reason: Not due yet (TRY_AT_EPOCH=${
+            lastResult.epoch
+          }, ${formatEpoch(lastResult.epoch)}). ${logOrderDetails}`,
           conditionalOrder.params
         );
         continue;
@@ -131,10 +133,14 @@ const _checkForAndPlaceOrder: ActionFn = async (
       // Check if the order is due (by blockNumber)
       if (
         lastResult?.result === PollResultCode.TRY_ON_BLOCK &&
-        lastResult.blockNumber < blockNumber
+        blockNumber < lastResult.blockNumber
       ) {
         console.log(
-          `${logPrefix} Skipping conditional. Reason: Not due yet (TRY_ON_BLOCK=${lastResult.blockNumber}). ${logOrderDetails}`,
+          `${logPrefix} Skipping conditional. Reason: Not due yet (TRY_ON_BLOCK=${
+            lastResult.blockNumber
+          }, in ${
+            lastResult.blockNumber - blockNumber
+          } blocks). ${logOrderDetails}`,
           conditionalOrder.params
         );
         continue;
--- a/actions/checkForAndPlaceOrder.ts
+++ b/actions/checkForAndPlaceOrder.ts
@@ -32,14 +32,14 @@ import {
 import { ChainContext, ConditionalOrder, OrderStatus } from "./model";
 import { pollConditionalOrder } from "./utils/poll";
 import {
+  OrderPostError,
   PollParams,
   PollResult,
   PollResultCode,
   PollResultErrors,
   PollResultSuccess,
-  PollResultTryNextBlock,
-  PollResultUnexpectedError,
   SupportedChainId,
+  formatEpoch,
 } from "@cowprotocol/cow-sdk";
 
 const GPV2SETTLEMENT = "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
@@ -53,6 +53,9 @@ const ORDER_BOOK_API_HANDLED_ERRORS = [
   "InsufficientAllowance",
   "InsufficientFee",
 ];
+
+const ApiErrors = OrderPostError.errorType;
+const WAITING_TIME_MINUTES_FOR_NOT_BALANCE = 10; // 10 min
 
 /**
  * Watch for new blocks and check for orders to place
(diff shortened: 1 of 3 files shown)


actions/checkForAndPlaceOrder.ts
```diff
@@ -116,10 +119,12 @@ const _checkForAndPlaceOrder: ActionFn = async (
       // Check if the order is due (by epoch)
       if (
         lastResult?.result === PollResultCode.TRY_AT_EPOCH &&
-        lastResult.epoch < blockTimestamp
+        blockTimestamp < lastResult.epoch
```
anxolin, line 122: This was flipped! 🙈

actions/checkForAndPlaceOrder.ts
```diff
@@ -128,10 +133,14 @@ const _checkForAndPlaceOrder: ActionFn = async (
       // Check if the order is due (by blockNumber)
       if (
         lastResult?.result === PollResultCode.TRY_ON_BLOCK &&
-        lastResult.blockNumber < blockNumber
+        blockNumber < lastResult.blockNumber
```
anxolin, line 136: this too! we really need unit testing and write this in a more functional approach to be able to test it easier

actions/checkForAndPlaceOrder.ts
```diff
@@ -423,16 +433,15 @@ export const _printUnfilledOrders = (orders: Map<BytesLike, OrderStatus>) => {
...
-  apiUrl: string,
-  orderRef: string
-): Promise<
-  | Omit<PollResultSuccess, "order" | "signature">
-  | PollResultTryNextBlock
-  | PollResultUnexpectedError
-> {
+async function _placeOrder(params: {
+  orderUid: string;
+  order: any;
```
alfetopito, line 438: Don't we have a type for the order?

actions/checkForAndPlaceOrder.ts
```diff
@@ -510,22 +524,43 @@ async function _placeOrder(
...
-  data: any,
-  error: any
-):
-  | Omit<PollResultSuccess, "order" | "signature">
-  | PollResultTryNextBlock
-  | PollResultUnexpectedError {
+function _handleOrderBookError(params: {
+  status: any;
+  data: any;
+  error: any;
```
alfetopito, line 530: So much `any` 🙈

actions/test/run_local.ts
```diff
@@ -52,9 +52,10 @@ const main = async () => {
       `[run_local] Processing ONCE for a specific transaction: ${txEnv}...`
     );
 
-    await processTx(provider, txEnv, chainId, testRuntime).catch(() =>
-      exit(ERROR_CODE_PROCESS_TX)
-    );
+    await processTx(provider, txEnv, chainId, testRuntime).catch((e) => {
+      console.error(e);
+      exit(ERROR_CODE_PROCESS_TX);
+    });
```
alfetopito, line 58: Can you use a try catch instead?

actions/test/run_local.ts
```diff
@@ -235,7 +236,7 @@ async function processTx(
 ) {
   let hasErrors = false;
   const transaction = await provider.getTransaction(tx);
-  if (!transaction.blockNumber) {
+  if (!transaction?.blockNumber) {
     throw new Error(`The transaction ${tx} is not mined yet (no blockNumber)`);
   }
   const block = await provider.getBlock(transaction.blockNumber);
```
alfetopito, line 242: I'll say this again.
Why do you need line 243?

It's already captured in line 239!

If you really want to check for the tx individually, move the check up

alfetopito (approved):
Nothing major, nitpicks mostly

actions/checkForAndPlaceOrder.ts
```diff
@@ -510,22 +524,43 @@ async function _placeOrder(
...
+      [
+        ApiErrors.INSUFFICIENT_ALLOWANCE,
+        ApiErrors.INSUFFICIENT_BALANCE,
+      ].includes(data?.errorType)
+    ) {
+      const nextPollTimestamp =
+        blockTimestamp + WAITING_TIME_MINUTES_FOR_NOT_BALANCE * 60;
+      return {
+        result: PollResultCode.TRY_AT_EPOCH,
+        epoch: blockTimestamp + WAITING_TIME_MINUTES_FOR_NOT_BALANCE,
```
alfetopito, line 555: Wait, this is not in seconds.
Why not use `nextPollTimestamp` here?

alfetopito: Also, maybe the const should be in seconds already to avoid this kind of issue?

anxolin: Not good reason, just call it i was dumb

actions/test/run_local.ts
anxolin: agree is not the best to finish execution in the middle. This part is being redone in the dockerised version.
I think is not worth it to through effort in this branch. We will need to improve a lot the standalone one.

actions/checkForAndPlaceOrder.ts
anxolin: this comes from other PRs, i hope is better in the new standalone

actions/test/run_local.ts
anxolin: 👍

actions/checkForAndPlaceOrder.ts
anxolin: we do, but this is from another PR and we will need to code freeze this until @mfw78 merges his waterfall.

alfetopito (approved):

PR #52, opened 2023-09-22
chore: dockerise

### LINKED ISSUE #51: chore: dockerise the service for deployment

mfw78:
# Description
For moving this service to production, it should be containerised so it can be deployed on our infra.

# Changes

- [x] Minor repo layout changes.
- [x] Dockerfile for building.
- [x] Check for settlement logic was removed (not being used) for build purposes.

## How to test

1. `docker build . -t watchtower`
2. `docker run -v ./database:/usr/src/app/database -it watchtower run --rpc http://172.33.0.6:8545 --rpc http://172.33.0.5:8545 --deployment-block 17883049 --deployment-block 29389123 --page-size 0 --one-shot true`

**NOTE**: The above assumes that you have a folder of `database` in the current directory which will be mapped to the docker image. Substitute the RPCs for your test RPCs.
**NOTE**: The order of the RPCs and deployment blocks must be the same for the respective chains.

## Related Issues

Closes #51


ahhda (approved):

package.json
```diff
@@ -39,6 +39,7 @@
     "typechain": "^8.1.1"
   },
   "dependencies": {
+    "@commander-js/extra-typings": "^11.0.0",
```
alfetopito, line 42: Why are the typings needed as dependencies rather than dev only?
I guess this packaged doesn't contain only typings then

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:alpine AS build
+
+WORKDIR /usr/src/app
+
+COPY . .
+
+RUN yarn && \
+    yarn build
+
+VOLUME [ "/usr/src/app/database" ]
```
alfetopito, line 10: Nitpick: would it be possible to use a parameterized path rather than hard coding it here?

alfetopito (approved):

.gitignore
```diff
@@ -14,4 +14,7 @@ node_modules/
 # Compiled production code
 /dist/
 .yalc
-yalc.lock
\ No newline at end of file
+yalc.lock
+
+# Databases
+database*
```
anxolin, line 20: you can ignore my comment in previous PRs

Why not `database/`

anxolin (approved):
Can you add the instructions in the README? 

Also, the instructions fail for me. I need to remove `./` from `./database` to make it work.

image

Other than that, it worked

.gitignore
mfw78: There may be different database directories. We should ignore them all :slightly_smiling_face:

package.json
mfw78: My mistake, I initially started going down the bundler rabithole but quickly aborted. Forgot to revert this - have now done so.

Dockerfile
mfw78: The correct place for this would likely best be the commander config options so the path was specifiable there. The volume in the dockerfile is hardcoded.

PR #53, opened 2023-09-22
refactor: context

### LINKED ISSUE #50: chore: standardise variable nomenclature
### Background
I'm frustrated by the creep that has been introduced over `chain`, `chainId`, `chainContext`, `context` etc. Let's standardise here :grinning:

### Details
Refactor the variables so that they have a clear meaning:

- `chainId` MUST be `SupportedChainId` | number | string
- `chainContext` MUST contain `chainId`, and `rpc`

### Acceptance criteria

- [ ] Clearly defined variables

mfw78:
# Description
Given that now the code has contexts which change across different chains, it's important to be careful about distinguishing this.

# Changes

- [x] Converged on using `ChainContext` class that contains all data for that chain.
- [x] Passes `this` down the code path for retrieval of providers, chainId etc.
- [x] Did some gardening removing storage back-off (now using storage with known transaction locking, so can fail, killing the container). 

## How to test

Confirm sync still works from genesis:

```bash
docker run -v ./database:/usr/src/app/database -it watchtower run --rpc http://172.33.0.6:8545 --rpc http://172.33.0.5:8545 --deployment-block 17883049 --deployment-block 29389123 --page-size 0 --one-shot true
```

## Related Issues

Closes #50

### DIFF
--- a/src/utils/context.ts
+++ b/src/utils/context.ts
@@ -1,5 +1,4 @@
 import Slack = require("node-slack");
-import { backOff } from "exponential-backoff";
 import DBService from "./db";
 
 import {
@@ -119,7 +118,7 @@ export async function handleExecutionError(e: any) {
 
     if (notified && executionContext) {
       executionContext.registry.lastNotifiedError = new Date();
-      await writeRegistry();
+      await executionContext.registry.write();
     }
   } catch (error) {
     console.error("Error sending slack notification", error);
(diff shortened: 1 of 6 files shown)


alfetopito (approved):

src/utils/context.ts
```diff
@@ -166,67 +165,6 @@ export function sendSlack(message: string): boolean {
...
-              `Error writing registry. Attempt ${attemptNumber}. Retrying...`,
-              e
-            );
-            return true;
-          },
-        }
-      ).catch((e) => {
-        if (executionContext) {
-          console.error(
-            "Error writing registry. Not more attempts! Dumping the orders",
```
anxolin, line 200: why removing the exponential-backoff logic? 

I believe there could still be some write error, right? don't we want to re-attempt?

mfw78: Write retry logic has been removed as there is an assumption now around it having atomic guarantees. If the storage fails, IMO the service should die, which will cause the container to cycle and this will become visible.

As the run loop now does a "warm up", there is no danger in it losing state as sync picks back up where it left off.

anxolin: fine, not a big deal, but even if the operations are atomic, the re-attempt will be also atomic, so no harm on re-attempting and trying to heal the system without another re-run

anyways, i don't think an IO operation on a docker volume will likely fail, so not a big deal. Its just, it was a small improvement and i didn't feel needed to be removed

Issue #18, opened 2023-08-29
refactor: model


anxolin:
Currently, watch tower depends on the model to do its tasks. 

Each version could introduce potentially breaking changes in the model, this of course would break the indexing. 
For this reason we kept the original model stable. 

However, it would be good to be able to evolve the model over time. 

Tasks:

- [x] Version model #21 
- [x] Do not use typechain generated structs

## Version model
This issue suggests using some versioning for the model, and some simple migration logics.

## Do not use IConditionalOrder.ConditionalOrderParamsStruct
IConditionalOrder.ConditionalOrderParamsStruct has a weird definition:

```ts
export type ConditionalOrderParamsStruct = {
    handler: PromiseOrValue<string>;
    salt: PromiseOrValue<BytesLike>;
    staticInput: PromiseOrValue<BytesLike>;
  };
```

Actually, this is not what ends up being persisted. See the hack: https://github.com/cowprotocol/tenderly-watch-tower/blob/88ab5365f303a0f316b70eaea98a1604767ba45a/actions/checkForAndPlaceOrder.ts#L201

```ts
const [handler, salt, staticInput] = conditionalOrder.params as any as [
      string,
      string,
      string
    ];
```

## 
https://github.com/cowprotocol/tenderly-watch-tower/blob/88ab5365f303a0f316b70eaea98a1604767ba45a/actions/checkForAndPlaceOrder.ts#L199

mfw78:
@anxolin Given #21 introduces versioning, I'll close this issue. If it hasn't been addressed, please re-open 🙏

mfw78:
Per #84 this now closes out the struct use.

PR #64, opened 2023-09-28
chore: orderbook sdk

### LINKED ISSUE #63: bug: becoming rate limited on orderbook api
### Problem
When using `axios` with no backoff / rate-limiter, in conjunction with a local ethereum node, the warm-up causes a large spike in API requests to the backend, resulting in rate limiting.

### Impact
This has a medium to high impact as it is likely to result in to-be-initialised watch-tower failing to warmup and sync to the current blockchain's state.

### To reproduce
1. Run locally to an ethereum / gnosis chain node.
2. Attempt to sync from the deployment block.
3. Observe rate limiting.

### Expected behaviour
Doesn't become rate limited.

### Tenderly watch-tower version/commit hash
- Commit hash `b5ae796`.

### Additional context
N/A

mfw78:
# Description
This PR addresses rate limiting when syncing with local (ie. high speed) nodes.

# Changes

- [x] Use `OrderBook` API from `cow-sdk`.
- [x] Remove `axios`.

## How to test

1. Run sync from deployment blocks and observe no failures:

```bash
yarn ts-node ./src/index.ts run --rpc http://10.23.0.2:8545 --rpc http://10.23.0.2:8645 --deployment-block 17883049 --deployment-block 29389123 --page-size 10000 > debug.txt 2>error.txt
```
## Related Issues

Fixes #63

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "@cowprotocol/cow-sdk": "^3.0.0-rc.9",
     "@sentry/integrations": "^7.64.0",
     "@sentry/node": "^7.64.0",
-    "axios": "^1.4.0",
     "ethers": "^5.7.2",
     "graphql-request": "^6.1.0",
     "level": "^8.0.0",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   "dependencies": {
     "@commander-js/extra-typings": "^11.0.0",
     "@cowprotocol/contracts": "^1.4.0",
-    "@cowprotocol/cow-sdk": "^3.0.0-rc.9",
+    "@cowprotocol/cow-sdk": "^3.0.0",
     "@sentry/integrations": "^7.64.0",
     "@sentry/node": "^7.64.0",
     "ethers": "^5.7.2",
--- a/src/domain/checkForAndPlaceOrder.ts
+++ b/src/domain/checkForAndPlaceOrder.ts
@@ -28,6 +28,7 @@ import {
 } from "../utils";
 import {
   OrderBookApi,
+  OrderCreation,
   OrderPostError,
   PollParams,
   PollResult,
@@ -436,7 +437,7 @@ async function _placeOrder(params: {
     params;
   const logPrefix = `[placeOrder::${orderRef}]`;
   try {
-    const postData = {
+    const postOrder: OrderCreation = {
       ...order,
       sellAmount: order.sellAmount.toString(),
       buyAmount: order.buyAmount.toString(),
@@ -444,13 +445,13 @@ async function _placeOrder(params: {
       signingScheme: "eip1271",
     };
 
-    // If the order is a dry run, don't post to the API
+    // If the operation is a dry run, don't post to the API
     console.log(
       `${logPrefix} Post order ${orderUid} to OrderBook on chain ${orderBook.context.chainId}`
     );
-    console.log(`${logPrefix} Order`, postData);
+    console.log(`${logPrefix} Order`, postOrder);
     if (!dryRun) {
-      const orderUid = await orderBook.sendOrder({ ...postData });
+      const orderUid = await orderBook.sendOrder(postOrder);
       console.log(`${logPrefix} API response`, { orderUid });
     }
   } catch (error: any) {
--- a/src/domain/checkForAndPlaceOrder.ts
+++ b/src/domain/checkForAndPlaceOrder.ts
@@ -5,8 +5,6 @@ import {
   computeOrderUid,
 } from "@cowprotocol/contracts";
 
-import axios from "axios";
-
 import { ethers } from "ethers";
 import { BytesLike, Logger } from "ethers/lib/utils";
 
@@ -29,6 +27,7 @@ import {
   pollConditionalOrder,
 } from "../utils";
 import {
+  OrderBookApi,
   OrderPostError,
   PollParams,
   PollResult,
@@ -266,7 +265,7 @@ async function _processConditionalOrder(
   context: ChainContext,
   orderRef: string
 ): Promise<PollResult> {
-  const { provider, apiUrl } = context;
+  const { provider, orderBook, dryRun } = context;
   try {
     // TODO: Fix model and delete the explicit cast: https://github.com/cowprotocol/tenderly-watch-tower/issues/18
     const [handler, salt, staticInput] = conditionalOrder.params as any as [
@@ -342,9 +341,10 @@ async function _processConditionalOrder(
       const placeOrderResult = await _placeOrder({
         orderUid,
         order: { ...orderToSubmit, from: owner, signature },
-        apiUrl,
+        orderBook,
         blockTimestamp,
         orderRef,
+        dryRun,
       });
 
       // In case of error, return early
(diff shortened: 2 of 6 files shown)


src/domain/checkForAndPlaceOrder.ts
```diff
@@ -424,55 +424,41 @@ export const _printUnfilledOrders = (orders: Map<BytesLike, OrderStatus>) => {
...
-      sellTokenBalance: order.sellTokenBalance,
-      buyTokenBalance: order.buyTokenBalance,
       signingScheme: "eip1271",
-      signature: order.signature,
-      from: order.from,
     };
 
-    // if the apiUrl doesn't contain localhost, post
-    console.log(`${logPrefix} Post order ${orderUid} to ${apiUrl}`);
+    // If the order is a dry run, don't post to the API
```
alfetopito, line 444: Did you mean:

```suggestion
    // If the operation is a dry run, don't post to the API
```

src/types/model.ts
```diff
@@ -132,9 +132,12 @@ export class Registry {
...
   ): Promise<Registry> {
     const db = storage.getDB();
-    const str = await db.get(
-      getNetworkStorageKey(CONDITIONAL_ORDER_REGISTRY_STORAGE_KEY, network)
-    );
+    const str = await db
+      .get(
+        getNetworkStorageKey(CONDITIONAL_ORDER_REGISTRY_STORAGE_KEY, network)
+      )
+      .then((str) => str)
```
alfetopito, line 139: Is this needed?

alfetopito (approved):

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -424,55 +424,41 @@ export const _printUnfilledOrders = (orders: Map<BytesLike, OrderStatus>) => {
...
+    if (!dryRun) {
+      const orderUid = await orderBook.sendOrder({ ...postData });
+      console.log(`${logPrefix} API response`, { orderUid });
     }
   } catch (error: any) {
     let reasonError = "Error placing order in API";
     if (error.response) {
-      const { status, data } = error.response;
+      const { status } = error.response;
+      const { body } = error;
```
anxolin, line 457: Has this been tested? does the SDK return the original response so we can get the status?

src/types/model.ts
```diff
@@ -132,9 +132,12 @@ export class Registry {
     genesisBlockNumber: number
   ): Promise<Registry> {
     const db = storage.getDB();
-    const str = await db.get(
-      getNetworkStorageKey(CONDITIONAL_ORDER_REGISTRY_STORAGE_KEY, network)
-    );
+    const str = await db
```
anxolin, line 135: not from this PR, but str is a not so good name for a var

src/types/model.ts
```diff
@@ -132,9 +132,12 @@ export class Registry {
...
+      .get(
+        getNetworkStorageKey(CONDITIONAL_ORDER_REGISTRY_STORAGE_KEY, network)
+      )
+      .then((str) => str)
+      .catch(() => undefined);
```
anxolin, line 140: Why do we hide ANY error here?

src/utils/misc.ts
```diff
@@ -1,37 +1,4 @@
...
-    case SupportedChainId.GOERLI:
-      return "https://api.cow.fi/goerli";
-    case SupportedChainId.GNOSIS_CHAIN:
-      return "https://api.cow.fi/xdai";
-    case LOCAL_CHAIN_ID:
-      return "http://localhost:3000";
-    default:
-      throw "Unsupported network";
-  }
-}
```
anxolin, line 33: happy to see you go!

package.json
```diff
@@ -44,7 +44,6 @@
     "@cowprotocol/cow-sdk": "^3.0.0-rc.9",
```
anxolin, line 44: you can use v3.0.0 now

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -424,55 +424,41 @@ export const _printUnfilledOrders = (orders: Map<BytesLike, OrderStatus>) => {
 async function _placeOrder(params: {
   orderUid: string;
   order: any;
```
anxolin, line 426: I think this order should have a type (and be build ourside of this method). 
The SDK should expose the different Order types, probably this one is the `OrderCreation`?

src/types/model.ts
mfw78: In the leveldb implementation, when using `get`, if the requested key doesn't exist, the function will throw a `404` (`LEVEL_NOT_FOUND`). Handling this case is required in the event of database initialisation.

src/types/model.ts
mfw78: At this point, the only condition that can be thrown by `get` is `LEVEL_NOT_FOUND`, when requesting a key that doesn't exist. This is the case on first run when populating from contract genesis. I see your point may be more about making this specific to the `LEVEL_NOT_FOUND` throw, and bubbling up all else, which is fair.

mfw78: Specifically handled not found (database initialisation), and bubble up the rest.

package.json
mfw78: Done.

src/types/model.ts
mfw78: Additionally, I've made the error handling here more specific to the initialisation case.

src/domain/checkForAndPlaceOrder.ts
mfw78: Included

src/domain/checkForAndPlaceOrder.ts
mfw78: I looked at this a bit, and I think we need to bump the `balanceToString`, `kindToString` methods etc from the watch tower to the `cow-sdk`. `kindToString` is already within the `composable` section, but it's not exported. I think we include this as part of #18

src/types/model.ts
mfw78: Will defer handling this to #65 as there are changes here that will result in more merge conflicts.

src/domain/checkForAndPlaceOrder.ts
mfw78: Yep, this has been tested :+1:

src/types/model.ts
alfetopito: I still don't get it.
You are talking about the error handling, but I'm asking about the `then((str) => str)`
If the key doesn't exist, it should not matter the `then` anyway.
Or is that required for some reason?

Also, I would prefer to use try/catch to prevent this sort of confusion, as Leveldb docs seem to prefer as well.

mfw78: I actually removed the `then`.

alfetopito: Not [on this PR](https://github.com/cowprotocol/watch-tower/pull/64/files#diff-b43f3ac95c01d64b54dbdfc80288236b116d9fe524c5e03dd74579f4de12edd5R139) 
image

Unless you mean another PR, which then explains things :)

mfw78: Ohh, sorry, too many PRs :see_no_evil:

src/domain/checkForAndPlaceOrder.ts
anxolin: yes, agree

PR #66, opened 2023-09-28
feat: watchdog

### LINKED ISSUE #54: bug: new block timeout
### Problem
There may be times where the RPC that the watch tower is connected to fails to advance the block chain. This is likely an RPC error, and should be reported via logging. Causes may include RPC rate limiting, connectivity issues, downtime.

### Impact
The impact of this is high. This would not result in _loss_ of funds, but would result in service degradation. 

### To reproduce

1. Run your own RPC.
2. Run the watch tower.
3. Disconnect your internet from your RPC :sweat_smile: 

### Expected behaviour
Logs should become noisy indicating that there is an underlying RPC issue.

mfw78:
# Description
This PR provides a main watch dog loop to both keep the block monitor running, and ensure that the chain continues to progress.

# Changes

- [x] Adds a watchdog loop to the block watcher

## How to test

1. Run normally using `yarn ts-node ./src/index.ts run --rpc http://10.23.0.2:8545 --rpc http://10.23.0.2:8645 --deployment-block 17883049 --deployment-block 29389123 --page-size 10000`.
2. Disconnect machine from the internet / network and observe watchdog timer counts up to 30s and then exits.

## Related Issues

Fixes #54

### DIFF
--- a/src/domain/chainContext.ts
+++ b/src/domain/chainContext.ts
@@ -10,6 +10,9 @@ import { checkForAndPlaceOrder } from "./checkForAndPlaceOrder";
 import { ethers } from "ethers";
 import { composableCowContract, DBService } from "../utils";
 
+const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
+const WATCHDOG_KILL_THRESHOLD = 30 * 1000; // 30 seconds
+
 /**
  * The chain context handles watching a single chain for new conditional orders
  * and executing them.
@@ -183,15 +186,23 @@ export class ChainContext {
    * 1. Check if there are any `ConditionalOrderCreated` events, and index these.
    * 2. Check if any orders want to create discrete orders.
    */
-  public async runBlockWatcher() {
-    const { provider, chainId } = this;
+  private async runBlockWatcher() {
+    const { provider, registry, chainId } = this;
     const _ = (s: string) =>
       console.log(`[runBlockWatcher:chainId:${chainId}] ${s}`);
     // Watch for new blocks
     _("Subscribe to new blocks");
+    let lastBlockReceived = 0;
+    let timeLastBlockProcessed = 0;
     provider.on("block", async (blockNumber: number) => {
       try {
         _(`New block ${blockNumber}`);
+        timeLastBlockProcessed = new Date().getTime();
+        if (blockNumber <= lastBlockReceived) {
+          // This may be a re-org, so process the block again
+          _(`Re-org detected, re-processing block ${blockNumber}`);
+        }
+        lastBlockReceived = blockNumber;
 
         const events = await pollContractForEvents(
           blockNumber,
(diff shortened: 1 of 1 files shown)


src/domain/chainContext.ts
```diff
@@ -209,11 +220,30 @@ export class ChainContext {
...
+      console.log(
+        `[runBlockWatcher:chainId:${chainId}] diff: ${timeElapsed}ms`
+      );
+
+      // If we haven't received a block in 30 seconds, restart the watcher
+      if (timeElapsed >= WATCHDOG_KILL_THRESHOLD) {
+        console.error(`[runBlockWatcher:chainId:${chainId}] Watchdog timeout`);
+        await registry.storage.close();
+        process.exit(1);
+      }
```
alfetopito, line 245: Comment doesn't match the code.
Auto-restarting might be a side effect, but the logic to restart it is not here.

src/domain/chainContext.ts
```diff
@@ -183,15 +186,23 @@ export class ChainContext {
...
     _("Subscribe to new blocks");
+    let lastBlockReceived = 0;
+    let timeLastBlockProcessed = new Date().getTime();
     provider.on("block", async (blockNumber: number) => {
       try {
         _(`New block ${blockNumber}`);
+        timeLastBlockProcessed = new Date().getTime();
+        if (blockNumber <= lastBlockReceived) {
+          // This may be a re-org, so process the block again
+          _(`Re-org detected, re-processing block ${blockNumber}`);
```
alfetopito, line 203: Still using `_` as a fn name?

alfetopito (approved):

src/domain/chainContext.ts
```diff
@@ -209,11 +220,30 @@ export class ChainContext {
...
+      // Todo: set this as trace to avoid spamming the logs
+      console.log(
+        `[runBlockWatcher:chainId:${chainId}] diff: ${timeElapsed}ms`
+      );
+
+      // If we haven't received a block in 30 seconds, restart the watcher
+      if (timeElapsed >= WATCHDOG_KILL_THRESHOLD) {
+        console.error(`[runBlockWatcher:chainId:${chainId}] Watchdog timeout`);
+        await registry.storage.close();
+        process.exit(1);
```
anxolin, line 244: nit: Maybe use a different error code than 1, just to signal this specific case.

src/domain/chainContext.ts
```diff
@@ -209,11 +220,30 @@ export class ChainContext {
...
+          `[runBlockWatcher:chainId:${chainId}] Error in pollContractForEvents for block ${blockNumber}`,
           error
         );
       }
     });
+
+    // We run a watchdog to check if we are receiving blocks.
+    while (true) {
+      // sleep for 5 seconds
+      await new Promise((resolve) => setTimeout(resolve, WATCHDOG_FREQUENCY));
```
anxolin, line 232: can you make a `sleep` util for this?

anxolin (approved):

src/domain/chainContext.ts
mfw78: Heh, yeah we have just settled on a debugging package and methodology, so all this is going to be given a good clean up in what will be a much anticipated PR :wink:

src/domain/chainContext.ts
mfw78: I don't think we should be getting into the business of custom error codes other than `0` or `1` as with this being dockerised, there's only a need to convey to docker that the process exited in a state other than normal.

src/domain/chainContext.ts
mfw78: Done.

src/domain/chainContext.ts
mfw78: Done.

PR #68, opened 2023-09-29
feat: unify logging

### LINKED ISSUE #67: chore: unify logging

mfw78:
# Description
Provides unification around logging levels and methodology given the movement to our cluster / ELK logging.

# Changes

- [x] Uses `loglevel`
- [x] Removes sentry / loggly

## How to test
Run and observe logging differences. If desiring granular logs, set environment variable `LOG_LEVEL` to 'DEBUG' or 'TRACE'.

## Related Issues

Fixes #67


src/domain/chainContext.ts
```diff
@@ -77,9 +77,9 @@ export class ChainContext {
    */
   public async warmUp(oneShot?: boolean) {
     const { provider, chainId } = this;
+    const log = logger.getLogger(`chainContext:warmUp:${chainId}`);
```
anxolin, line 80: The chain will be relevant for any log, but only be visible in this logger (other loggers will loose this info).

If we stick to one network per instance, simplifies our logging so we don't need to include the chain id.
Since it still supports multiple RPC nodes, its fine to stay, but as i mentioned, i don't think is a good practice (unless you overcomplicate the other loggers to include the chainId, among other changes that this code would require to safely run on multi-network setup)

src/domain/chainContext.ts
```diff
@@ -97,16 +97,18 @@ export class ChainContext {
           currentBlockNumber = await provider.getBlockNumber();
           toBlock = toBlock > currentBlockNumber ? currentBlockNumber : toBlock;
 
-          _(
+          log.trace(
```
anxolin, line 100: THANKSSSSSSSSSSSSSS, so nice to see `_` gone

src/domain/chainContext.ts
```diff
@@ -97,16 +97,18 @@ export class ChainContext {
...
             `Reaching tip of chain, current block number: ${currentBlockNumber}`
           );
         }
 
-        _(`Processing events from block ${fromBlock} to block ${toBlock}`);
+        log.trace(
+          `Processing events from block ${fromBlock} to block ${toBlock}`
+        );
```
anxolin, line 107: Nice that this is TRACE level!

src/domain/chainContext.ts
```diff
@@ -232,14 +233,12 @@ export class ChainContext {
       await asyncSleep(WATCHDOG_FREQUENCY);
       const currentTime = new Date().getTime();
       const timeElapsed = currentTime - timeLastBlockProcessed;
-      // Todo: set this as trace to avoid spamming the logs
-      console.log(
-        `[runBlockWatcher:chainId:${chainId}] diff: ${timeElapsed}ms`
-      );
+
+      log.trace(`diff: ${timeElapsed}ms`);
```
anxolin, line 237: nit: Can you add more context to the log?

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -138,8 +143,8 @@ async function _checkForAndPlaceOrder(
         lastHint?.result === PollResultCode.TRY_ON_BLOCK &&
         blockNumber < lastHint.blockNumber
       ) {
-        console.log(
-          `${logPrefix} Skipping conditional. Reason: Not due yet (TRY_ON_BLOCK=${
+        log.info(
```
anxolin, line 146: I would change this to `debug`, this will eliminate a lot of logs. Same for the one above. 
DCA trading once a month will log on every block we are skipping the order.

src/utils/context.ts
```diff
@@ -72,44 +58,8 @@ function _getSlack(options: SingularRunOptions): Slack | undefined {
...
-    });
-  }
-
-  // Return transaction
-  return sentryStartTransaction({
-    name: transactionName,
-    op: "action",
-  });
-}
-
```
anxolin, line 111: not from this PR, but good call removing

src/utils/logging.ts
```diff
@@ -1,98 +1,25 @@
...
+import * as prefix from "loglevel-plugin-prefix";
+import { Chalk } from "chalk";
+import chalk = require("chalk");
+
+const colors: Record<string, Chalk> = {
+  TRACE: chalk.magenta,
+  DEBUG: chalk.cyan,
+  INFO: chalk.blue,
+  WARN: chalk.yellow,
+  ERROR: chalk.red,
```
anxolin, line 11: Fancy :)

anxolin (approved):
Beautiful!

anxolin:
@mfw78 I think it would be great to provide some instructions in the README on how to increase/decrease logging verbosity using an ENV

Also, it might be my setup, but this is how it looks in my terminal (unreadable timestamp)

image

Also the format of the time is strange

src/commands/dumpDb.ts
```diff
@@ -1,17 +1,18 @@
 import { DumpDbOptions, Registry } from "../types";
-import { DBService } from "../utils";
+import { logger, DBService } from "../utils";
 
 /**
  * Dump the database as JSON to STDOUT for a given chain ID
  * @param options A dict, but essentially just the chainId
  */
 export async function dumpDb(options: DumpDbOptions) {
+  const log = logger.getLogger("commands:dumpDb");
```
alfetopito, line 9: Would it make sense to create the log instance in the module scope instead or getting it every time the fn runs?

Might not matter than much, actually.

src/utils/logging.ts
```diff
@@ -1,98 +1,25 @@
...
-  } else if (data.length === 1) {
-    return { level, message: data[0] };
-  }
-
-  return undefined;
-}
+import * as log from "loglevel";
+import * as prefix from "loglevel-plugin-prefix";
+import { Chalk } from "chalk";
+import chalk = require("chalk");
```
alfetopito, line 4: 2 types of imports for the same lib?

src/index.ts
```diff
@@ -1,6 +1,16 @@
 import { program, Option } from "@commander-js/extra-typings";
 import { ReplayTxOptions } from "./types";
 import { dumpDb, replayBlock, replayTx, run } from "./commands";
+import { logger } from "./utils";
+
+type LogLevelNames = {
+  [K in keyof typeof logger.levels]: K;
+};
```
alfetopito, line 8: I feel like this should live within logger utils and also be used there on `colors` type for example.

alfetopito (approved):
Nice!

Some nitpicks

src/index.ts
mfw78: This has been addressed by @anxolin in #69

src/domain/chainContext.ts
mfw78: You're welcome. `_` is retired on a beach drinking cocktails.

src/domain/chainContext.ts
mfw78: As discussed, will retain multi-chain capabilities (for a dappnode package). Loggers are injecting the `chainId`.

src/domain/chainContext.ts
mfw78: Done

src/domain/checkForAndPlaceOrder.ts
mfw78: Done.

src/utils/logging.ts
mfw78: This was related to not have `esModuleInterop` set to `true` in the `tsconfig.json`. Or some weird error here. JS is a mystery to me.

mfw78:
> @mfw78 I think it would be great to provide some instructions in the README on how to increase/decrease logging verbosity using an ENV
> 
> Also, it might be my setup, but this is how it looks in my terminal (unreadable timestamp)
> image
> 
> Also the format of the time is strange

FYI, this has been addressed as well in this PR.

PR #74, opened 2023-10-02
feat: on-chain polling hints

### LINKED ISSUE #73: feat: smart contract hinted polling
### Problem
It was always frustrating with the initial orders from `ComposableCoW` not being able to relay important information to the watch-tower to allow it to optimise it's polling / performance.

### Suggested solution
Custom errors returned from `ComposableCoW` have been expanded given the learnings so as to provide hints to the watch-tower. These should be integrated and used in the `pollLegacy` logic.

### Alternatives considered
Implement custom logic for each handler (order type). This is not a scalable and resilient solution though as it requires all conditional order authors to duplicate business logic between the smart contract and an SDK module. This then also means that they need to PR / push upstream this logic for the watch-tower to consume it.

### Acceptance criteria

- [x] Conditional orders are able to provide hints to the watch-tower.

mfw78:
# Description
In order to address on-chain polling hints, `ComposableCoW` interfaces introduced some additional custom error reversions. This supports these to enable enhanced on-chain hinting to the watch-tower.

# Changes

- [x] Removes legacy (nethermind and non-nethermind code) to use a unified RPC for custom error handling.
- [x] Adds handling for all custom error handling and revert messages.
- [x] Refactors / increases test coverage

## How to test

1. `yarn test`

## Related Issues

Fixes #73


shoom3301 (approved):

src/domain/chainContext.ts
```diff
@@ -13,6 +16,8 @@ import { composableCowContract, DBService, getLogger } from "../utils";
 const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
 const WATCHDOG_KILL_THRESHOLD = 30 * 1000; // 30 seconds
 
+const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
```
anxolin, line 19: I think this logic should be moved eventually to the SDK

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -579,7 +556,8 @@ async function _pollLegacy(
 
     const [{ success, returnData }] = lowLevelCall;
 
-    // If the call failed, we throw an error and wrap the returnData as it is done by erigon / geth 🦦
+    // If the call failed, there may be a custom error to provide hints. We wrap the error in a LowLevelError
+    // so that it can be handled in the catch.
     if (!success) {
       throw new LowLevelError("low-level call failed", returnData);
```
anxolin, line 562: Both the error and the message seems to generic. It looks very coupled to `multical aggregate3` call, also the message should probably mention it failed the `getTradeableOrderWithSignature`. Lastly, it might be a good idea to print the Tenderly simulation as a ERROR instead of DEBUG in case of a failure. It might make debugging easier (since `DEBUG` level will likely be disabled in production)

src/utils/contracts.ts
```diff
@@ -11,11 +11,32 @@ const REQUIRED_SELECTORS = [
...
+  INTERFACE_NOT_SUPPORTED: "InterfaceNotSupported()",
+  ORDER_NOT_VALID: "OrderNotValid(string)",
+  POLL_TRY_NEXT_BLOCK: "PollTryNextBlock(string)",
+  POLL_NEVER: "PollNever(string)",
+  POLL_TRY_AT_BLOCK: "PollTryAtBlock(uint256,string)",
+  POLL_TRY_AT_EPOCH: "PollTryAtEpoch(uint256,string)",
+};
+
+// Just export the keys of the CUSTOM_ERROR_ABI_MAP as a type
+export type CustomError = keyof typeof CUSTOM_ERROR_ABI_MAP;
```
anxolin, line 29: This error name is too generic

src/utils/utils.spec.ts
```diff
@@ -40,128 +35,66 @@ describe("test against concrete examples", () => {
...
+      blockNumberOrEpoch: 1694340000,
+    });
   });
 
-  it("should pass the OrderNotValid selector correctly from nethermind", () => {
-    expect(parseCustomError(ORDER_NOT_VALID_NETHERMIND)).toMatchObject({
-      errorNameOrSelector: ORDER_NOT_VALID_SELECTOR,
+  it("should pass the PollNever selector correctly", () => {
+    expect(customErrorDecode(POLL_NEVER)).toMatchObject({
+      selector: "POLL_NEVER",
```
anxolin, line 77: cool tests!

package.json
```diff
@@ -35,7 +35,7 @@
     "lint-staged": "^14.0.1",
     "prettier": "^2.8.8",
     "ts-node": "^10.9.1",
-    "typechain": "^8.1.1"
+    "typechain": "^8.3.1"
```
anxolin, line 38: will the autogenerated types change? if so, maybe a follow up to re-create the types might be benefitial

alfetopito (approved):

anxolin:
@mfw78 I rebased and resolved the conflicts. You can still address comments if you want as follow up. I just need to consolidate all https://github.com/cowprotocol/watch-tower/issues/89

package.json
mfw78: Autogenerated types aren't committed into the repository so there's nothing to recreate. This is actually just a version bump that didn't result in any substantive change (I was hopeful that this was going to include custom error types, but it seems this is [still yet to be done in typechain](https://github.com/dethcrypto/TypeChain/issues/667).

src/domain/checkForAndPlaceOrder.ts
mfw78: The `Multicall3.aggregate3Value`, literally conducts a low-level solidity calls (hence the `LowLevelError` name). 

In solidity, you can do a `.call` and this allows the EVM to provide success (or failure, ie. revert), and the return data from the EVM call. It's most definitely very coupled, but contextually, there's only one place that this happens.

src/domain/chainContext.ts
mfw78: Eventually, maybe. But we need a place to point the Valory guys to for a commit hash, which we have now, for what Valory should be doing :slightly_smiling_face:

PR #94, opened 2023-10-05
fix: drop incompatible orders

### LINKED ISSUE #91: bug: low level calls cause log spamming if not adhering to the interface
### Problem
When the watch tower polls a custom order type, the selector / revert error returned has no _guarantee_ of being compliant with the interface. It's currently being observed that old TWAP orders from the pre-production version of the contracts are returning `OrderNotValid()` selectors, as opposed to `OrderNotValid(string)` which is causing these to not be recognised by the low-level call handlers.

### Impact
Causes needless log spamming, and doesn't promote strict interface adherence. 

### To reproduce
1. Observe the goerli watch tower in production.

### Expected behaviour
If an order doesn't adhere to the interface, it should **not** be monitored. Any order that doesn't return a known custom error for the revert should be dropped immediately.

### Screenshots/logs

```raw
2023-10-05T01:31:28.999Z ERROR checkForAndPlaceOrder:_pollLegacy:5:2.8@9810859: Error on CALL to getTradeableOrderWithSignature. Simulate: https://dashboard.tenderly.co/gp-v2/watch-tower-prod/simulator/new?network=5&contractAddress=0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74&rawFunctionInput=0x26e0a1960000000000000000000000008654d1136f2a760ba3e1c9e131cb9ad217921b52000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000002400000000000000000000000000000000000000000000000000000000000000260000000000000000000000000910d00a310f7dc5b29fe73458f47f519be547d3d000000000000000000000000000000000000000000000000000000189e3ae4af00000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000140000000000000000000000000b4fbf271143f4fbf7b91a5ded31805e42b2208d600000000000000000000000091056d4a53e1faa1a84306d4deaec71085394bc80000000000000000000000008654d1136f2a760ba3e1c9e131cb9ad217921b5200000000000000000000000000000000000000000000000000b1a2bc2ec500000000000000000000000000000000000000000000000000062a37aa81e9beb9780000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000007080000000000000000000000000000000000000000000000000000000000000000506960793899dbd9225c61c44bd44927462151c82dec95315cf937f6b95ef21f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
2023-10-05T01:31:28.999Z ERROR checkForAndPlaceOrder:_handleGetTradableOrderCall:5:2.8@9810859: checkForAndPlaceOrder:_handleGetTradableOrderCall:5:2.8@9810859 Unexpected error LowLevelError: low-level call failed
    at _pollLegacy (/usr/src/app/dist/src/domain/checkForAndPlaceOrder.js:340:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async _processConditionalOrder (/usr/src/app/dist/src/domain/checkForAndPlaceOrder.js:141:26)
    at async checkForAndPlaceOrder (/usr/src/app/dist/src/domain/checkForAndPlaceOrder.js:70:32)
    at async processBlock (/usr/src/app/dist/src/domain/chainContext.js:213:20)
    at async JsonRpcProvider.<anonymous> (/usr/src/app/dist/src/domain/chainContext.js:155:21) {
  data: '0xf3ec7a9f'
}
```

In this case, `0xf3ec7a9f` corresponds to the previous `OrderNotValid()` selector prior to the insertion of the `string` for there reason (implemented just before push to production).

### Tenderly watch-tower version/commit hash

Version: `v1.0.1-rc.0`

mfw78:
# Description
Currently incompatible orders (where there is a low level reversion that's not expected) are *NOT* deleted. This corrects that.

# Changes

- [x] Drop orders that return non-compatible interface revert messages.

## How to test
1. Sync goerli from genesis and confirm no unexpected errors.

## Related Issues

Fixes #91

### DIFF
--- a/src/domain/checkForAndPlaceOrder.ts
+++ b/src/domain/checkForAndPlaceOrder.ts
@@ -510,11 +510,10 @@ async function _pollLegacy(
   offchainInput: string,
   orderRef: string
 ): Promise<PollResult> {
-  const { chainId, contract, multicall } = context;
+  const { contract, multicall } = context;
   // as we going to use multicall, with `aggregate3Value`, there is no need to do any simulation as the
   // calls are guaranteed to pass, and will return the results, or the reversion within the ABI-encoded data.
   // By not using `populateTransaction`, we avoid an `eth_estimateGas` RPC call.
-  const log = getLogger(`checkForAndPlaceOrder:_pollLegacy:${orderRef}`);
   const to = contract.address;
   const data = contract.interface.encodeFunctionData(
     "getTradeableOrderWithSignature",
@@ -550,22 +549,29 @@ async function _pollLegacy(
       signature,
     };
   } catch (error: any) {
-    log.error(
-      `Error on CALL to getTradeableOrderWithSignature. Simulate: https://dashboard.tenderly.co/gp-v2/watch-tower-prod/simulator/new?network=${chainId}&contractAddress=${to}&rawFunctionInput=${data}`
-    );
-
     // An error of some type occurred. It may or may not be a hint. We pass it to the handler to decide.
-    return _handleGetTradableOrderWithSignatureCall(error, owner, orderRef);
+    return _handleGetTradableOrderWithSignatureCall(
+      error,
+      owner,
+      orderRef,
+      context,
+      to,
+      data
+    );
   }
 }
 
 function _handleGetTradableOrderWithSignatureCall(
   error: any,
   owner: string,
-  orderRef: string
+  orderRef: string,
+  context: ChainContext,
+  to: string,
+  data: string
 ): PollResultErrors {
   const logPrefix = `checkForAndPlaceOrder:_handleGetTradableOrderCall:${orderRef}`;
   const log = getLogger(logPrefix);
+  const { chainId } = context;
 
   // If the error is a LowLevelError, we extract the selector, and any parameters.
   if (error instanceof LowLevelError) {
@@ -659,19 +665,24 @@ function _handleGetTradableOrderWithSignatureCall(
           };
       }
     } catch (err: any) {
-      log.error(`${logPrefix} Unexpected error`, err);
+      // Any errors thrown here can _ONLY_ come from non-compliant interfaces (ie. bad revert ABI encoding).
+      // We log the error, and return a DONT_TRY_AGAIN result.
+      // TODO: Add metrics to track this
+      log.error(
+        `Contract returned non-interface compliant revert via getTradeableOrderWithSignature. Simulate: https://dashboard.tenderly.co/gp-v2/watch-tower-prod/simulator/new?network=${chainId}&contractAddress=${to}&rawFunctionInput=${data}`
+      );
       return {
-        result: PollResultCode.UNEXPECTED_ERROR,
+        result: PollResultCode.DONT_TRY_AGAIN,
         reason:
-          "UnexpectedErrorName: Unexpected error" +
+          "Non-compliant interface error" +
           (err.message ? `: ${err.message}` : ""),
-        error: err,
       };
     }
   }
 
   log.error(`${logPrefix} ethers/call Unexpected error`, error);
-  // If we don't know the reason, is better to not delete the order
+  // We can only get here from some provider / ethers failure. As the contract hasn't had it's say
+  // we will defer to try again.
   // TODO: Add metrics to track this
   return {
     result: PollResultCode.TRY_NEXT_BLOCK,
--- a/src/domain/checkForAndPlaceOrder.ts
+++ b/src/domain/checkForAndPlaceOrder.ts
@@ -10,7 +10,6 @@ import { BytesLike } from "ethers/lib/utils";
 
 import { ConditionalOrder, OrderStatus } from "../types";
 import {
-  LowLevelError,
   formatStatus,
   getLogger,
   pollConditionalOrder,
@@ -511,6 +510,8 @@ async function _pollLegacy(
   orderRef: string
 ): Promise<PollResult> {
   const { contract, multicall } = context;
+  const logPrefix = `checkForAndPlaceOrder:_pollLegacy:${orderRef}`;
+  const log = getLogger(logPrefix);
   // as we going to use multicall, with `aggregate3Value`, there is no need to do any simulation as the
   // calls are guaranteed to pass, and will return the results, or the reversion within the ABI-encoded data.
   // By not using `populateTransaction`, we avoid an `eth_estimateGas` RPC call.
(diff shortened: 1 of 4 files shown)


alfetopito (approved):

anxolin (approved):

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -550,22 +549,29 @@ async function _pollLegacy(
       signature,
     };
   } catch (error: any) {
-    log.error(
```
anxolin, line 553: not from this PR, but, i think the `low-level call failed` log message is insufficient. It should be giving a context the call was `getTradeableOrderWithSignature`

anxolin (approved):

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -659,19 +665,24 @@ function _handleGetTradableOrderWithSignatureCall(
...
-      log.error(`${logPrefix} Unexpected error`, err);
+      // Any errors thrown here can _ONLY_ come from non-compliant interfaces (ie. bad revert ABI encoding).
+      // We log the error, and return a DONT_TRY_AGAIN result.
+      // TODO: Add metrics to track this
+      log.error(
+        `Contract returned non-interface compliant revert via getTradeableOrderWithSignature. Simulate: https://dashboard.tenderly.co/gp-v2/watch-tower-prod/simulator/new?network=${chainId}&contractAddress=${to}&rawFunctionInput=${data}`
+      );
       return {
-        result: PollResultCode.UNEXPECTED_ERROR,
+        result: PollResultCode.DONT_TRY_AGAIN,
```
anxolin, line 675: I don't think this is a good idea. catching ANY error thrown in the block above and just stop indexing the order might be a source of bugs.

Can't we handle it in a different way? 

Maybe could be usefult to extract the error handling to a testable function and create some tests. This way, we can make sure we capture the cases you want to capture and there's no future regressions. 

Also, by making the error handling independent, it will make `checkForAndPlaceOrders` thinner, which would be a good thing


Alternativelly, we could move the whole legacy polling and do the same. This in fact is something that is good is independent, since it should be moved to the SDK. Watch Tower logic should be simpler

src/domain/checkForAndPlaceOrder.ts
anxolin: okokko

src/domain/checkForAndPlaceOrder.ts
anxolin: > The only code path here (yes, it should be tested) that should throw and result in invalidating the order, is customErrorDecode that solely depends on only ABI decoding

I would hope we structure it so typescript already can check this assumption 

> Comprehensive logging. We have ELK in place, and much work has been done with logging to ensure great visibility.
Comprehensive metrics. In https://github.com/cowprotocol/watch-tower/pull/95 the order tuple (chainId, handler, owner, id) is labelled against the watch_tower_polling_onchain_invalid_interface_total metric. Therefore we can see when orders are getting dropped due to this, and investigate to ensure compliance.

I think these don't solve the problem. I think it should be impossible to remove an order if there's some other error that is not the decoding. Typescript can probably help with these at compile time

> We do some unit testing - https://github.com/cowprotocol/watch-tower/issues/97 - and validate while in prod.

This will help indeed

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -533,152 +533,155 @@ async function _pollLegacy(
...
-            `${selector}: Order for safe ${owner} where the Safe has swap guard enabled. Deleting order...`
+  try {
+    // The below will throw if:
+    // - the error is not a custom error (ie. the selector is not in the map)
+    // - the error is a custom error, but the parameters are not as expected
+    const { selector, message, blockNumberOrEpoch } =
+      customErrorDecode(returnData);
+    switch (selector) {
+      case "SINGLE_ORDER_NOT_AUTHED":
+      case "PROOF_NOT_AUTHED":
```
anxolin, line 592: We have this literal in 3 points in the app. Should we make an enum or sth to make sure typescript prevents us from making mistakes and we can keep the code dry?

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -533,152 +533,155 @@ async function _pollLegacy(
...
+  } finally {
+    // We use finally here to ensure that we always log the error, even if we hit an unexpected error
+    // Any errors thrown here can _ONLY_ come from non-compliant interfaces (ie. bad revert ABI encoding).
+    // We log the error, and return a DONT_TRY_AGAIN result.
+    // TODO: Add metrics to track this
+    log.debug(
+      `Contract returned non-interface compliant revert via getTradeableOrderWithSignature. Simulate: https://dashboard.tenderly.co/gp-v2/watch-tower-prod/simulator/new?network=${chainId}&contractAddress=${to}&rawFunctionInput=${data}`
+    );
+    return {
+      result: PollResultCode.DONT_TRY_AGAIN,
```
anxolin, line 681: because the `case` is not exaustive, we can still end up here. Is this inteded we don't try again for unknown errors?

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -533,152 +533,155 @@ async function _pollLegacy(
...
+      reason: "Non-compliant interface",
```
anxolin, line 682: nit: good to add a comment pointing to the interface they should implement, or what non-compliant means here

src/domain/checkForAndPlaceOrder.ts
mfw78: The switch is now exhaustive with the use of typescript enum's / types. Turns out I was missing one! :see_no_evil:

src/domain/checkForAndPlaceOrder.ts
mfw78: Enums have been implemented in typescript, and custom error handler has been moved to `contract.ts` within `utils`.

src/domain/checkForAndPlaceOrder.ts
mfw78: With the new typescript enum / types for the custom errors, the switch statement is now exhaustive.

src/domain/checkForAndPlaceOrder.ts
mfw78: Done.

src/domain/checkForAndPlaceOrder.ts
mfw78: I've gone through and implemented everything with very strong typing now to enforce the assumptions. Some tests have been added as well.

src/utils/contracts.ts
```diff
@@ -1,42 +1,82 @@
...
+        | CustomErrorSelectors.ORDER_NOT_VALID
+        | CustomErrorSelectors.POLL_TRY_NEXT_BLOCK
+        | CustomErrorSelectors.POLL_NEVER
+    ? { selector: K; message: string }
+    : K extends
+        | CustomErrorSelectors.POLL_TRY_AT_BLOCK
+        | CustomErrorSelectors.POLL_TRY_AT_EPOCH
+    ? { selector: K; message: string; blockNumberOrEpoch: number }
+    : never;
+}[CustomErrorSelectors];
```
alfetopito, line 52: 🤯

alfetopito (approved):

src/domain/checkForAndPlaceOrder.ts
anxolin: Looks better, but you are still not exhaustive. You can't be. Because you use `parsedCustomError.selector` in the switch, you can't know all selectors. You need to handle an unknown selector, but your code now assumes its impossible that will happen

src/domain/checkForAndPlaceOrder.ts
anxolin: I still don't love this error handling. We have a looooong TRY, and if there's ANY error, then we just discard the order

I feel you are making code assumptions, like "the only error would happen in `parsedCustomError.selector` and because of a `non-compliant` order, but i think its easy your assumption changes (i.e we add a future change during that looong TRY or in `parseCustomError` that throws an error. Then we will drop the order when there's no need,  and it will be hard to catch these cases (it will  happen silently)

src/utils/contracts.ts
```diff
@@ -1,42 +1,82 @@
...
-);
+export const abiToSelector = (abi: string): string =>
+  ethers.utils.id(abi).slice(0, 10);
+
+for (const errorType of Object.keys(
+  CUSTOM_ERROR_ABI_MAP
+) as CustomErrorSelectors[]) {
+  const selector = abiToSelector(CUSTOM_ERROR_ABI_MAP[errorType]);
+  CUSTOM_ERROR_SELECTOR_MAP[selector] = errorType;
+}
```
anxolin, line 79: I guess this is a init code. it wiuld be more clear if you do a IIFE to asign the inital value to `CUSTOM_ERROR_SELECTOR_MAP` or you create a named function that gets the value and you assign it to it.

src/utils/contracts.ts
```diff
@@ -72,64 +112,184 @@ export function composableCowContract(
...
+        // uint32. It is therefore safe to throw if the value is outside of that
+        // for javascript's number type.
+        if (blockNumberOrEpoch.gt(MAX_UINT32)) {
+          throw new Error("Block number or epoch out of bounds");
+        }
+
+        return {
+          selector,
+          message: msg as string,
+          blockNumberOrEpoch: (blockNumberOrEpoch as BigNumber).toNumber(),
```
anxolin, line 180: do we really need this hard casting of types?

src/utils/contracts.ts
```diff
@@ -72,64 +112,184 @@ export function composableCowContract(
...
+    );
+    const iface = new ethers.utils.Interface([fragment]);
+
+    switch (selector) {
+      case CustomErrorSelectors.PROOF_NOT_AUTHED:
+      case CustomErrorSelectors.SINGLE_ORDER_NOT_AUTHED:
+      case CustomErrorSelectors.INTERFACE_NOT_SUPPORTED:
+      case CustomErrorSelectors.INVALID_FALLBACK_HANDLER:
+      case CustomErrorSelectors.INVALID_HANDLER:
+      case CustomErrorSelectors.SWAP_GUARD_RESTRICTED:
```
anxolin, line 156: great you are using enum now! 😊

src/utils/contracts.ts
anxolin: Agree, this feels too complicated. What is this type needed for? 

Why not the function just returns a `PollResultErrors` so you also simplify the othe switch.

mfw78: This type covers the exact interface for all `ComposableCoW`-compatible orders (as returned from ABI on-chain). We control this interface, and therefore we can assert that _only_ these revert types are valid. If an order fails to adhere to this interface, it is to be dropped.

Typechain _does not_ do generation of typings for solidity custom error reverts (despite them being in the ABI), and therefore, we've basically had to implement this ourselves, which is essentially what this type is.

src/utils/contracts.ts
mfw78: I moved the explicit hard casting to be on the same line as the `decode`. The hard casting is required because when ABI decoding in ethers v5, the return type is `Result`, which has the type:

```typescript
export interface Result extends ReadonlyArray<any> {
    readonly [key: string]: any;
}
```

IMO we should always hard cast from a decode if the decode is returning `Result`, as this increases the readability of the code, and enforces that Typescript catches more potential issues.

src/utils/contracts.ts
```diff
@@ -72,64 +112,184 @@ export function composableCowContract(
...
+        }
+
+        return {
+          selector,
+          message: msg as string,
+          blockNumberOrEpoch: (blockNumberOrEpoch as BigNumber).toNumber(),
+        };
     }
+  } catch (err) {
+    throw err;
```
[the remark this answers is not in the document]
mfw78: Great catch, it isn't needed - consider it gone!

src/domain/checkForAndPlaceOrder.ts
mfw78: The revert data is parsed in `parseCustomError` to ensure that it complies with the ComposableCoW interface. If it does not, it will throw. Therefore the switch is exhaustive as only valid selectors reach that point in the execution path.

Invalid selectors (ie. that do _not_ comply with the interface that we have prescribed), are caught, dropped, and metrics are in place to monitor this.

src/utils/contracts.ts
mfw78: Done.

PR #96, opened 2023-10-06
chore: health probe

### LINKED ISSUE #70: feat: observability
### Problem
As the standalone watch-tower moves into production, observability is critical in ensuring the reliability of the service.

### Suggested solution
Introduce an express API that provides health & readiness monitoring, in addition to a metrics collection endpoint for automatic prometheus scraping.

### Alternatives considered
Use of loggly / sentry. These have been trialed, and while loggly is suitable for some log analysis, this can easily be achieved by ELK. Loggly as well doesn't provide much in terms of metrics monitoring - this is the bread and butter of Prometheus / Grafana.

### Acceptance criteria

- [x] `/health` endpoint that returns not ready if the chain is still in warm-up.
- [x] `/metrics` endpoint with targeted metrics distributed throughout.

mfw78:
# Description
This PR provides overall health monitoring for the block watcher (chain contexts). This assists with Kubernetes deployment to enable health probes for automatic restarting and monitoring.

# Changes

- [x] All chain contexts added to a static mapping for use in other services (ie. API).
- [x] Implemented `/health` API endpoint, returning >= 400 when the chain is not synced (in warm-up).

## How to test

1. Start syncing chain from scratch.
2. Observe the `/health` API endpoint returns `500` status error, with JSON showing current chain context status.
3. Once in sync, observe the `/health` API endpoint returns `200` status with JSON showing the current chain status.

## Related Issues

Fixes #70

### DIFF
--- a/src/domain/chainContext.ts
+++ b/src/domain/chainContext.ts
@@ -6,12 +6,19 @@ import {
   Multicall3,
   ComposableCoW,
   Multicall3__factory,
+  RegistryBlock,
+  blockToRegistryBlock,
 } from "../types";
 import { SupportedChainId, OrderBookApi } from "@cowprotocol/cow-sdk";
 import { addContract } from "./addContract";
 import { checkForAndPlaceOrder } from "./checkForAndPlaceOrder";
 import { ethers } from "ethers";
-import { composableCowContract, DBService, getLogger } from "../utils";
+import {
+  composableCowContract,
+  DBService,
+  getLogger,
+  isRunningInKubernetesPod,
+} from "../utils";
 import {
   blockHeight,
   blockProducingRate,
@@ -26,8 +33,12 @@ const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
 const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
 
 enum ChainSync {
-  SYNCED = "SYNCED",
+  /** The chain is currently in the warm-up phase, synchronising from contract genesis or lastBlockProcessed */
   SYNCING = "SYNCING",
+  /** The chain is in sync with the latest block */
+  IN_SYNC = "IN_SYNC",
+  /** The chain is in an unknown state based on duration of time since a block was processed */
+  UNKNOWN = "UNKNOWN",
 }
 
 type Chains = { [chainId: number]: ChainContext };
@@ -35,7 +46,7 @@ type Chains = { [chainId: number]: ChainContext };
 export interface ChainStatus {
   sync: ChainSync;
   chainId: SupportedChainId;
-  lastProcessedBlock: number;
+  lastProcessedBlock: RegistryBlock | null;
 }
 
 export interface ChainHealth extends ChainStatus {
@@ -129,13 +140,13 @@ export class ChainContext {
     const { pageSize } = this;
 
     // Set the block height metric
-    blockHeight.labels(chainId.toString()).set(lastProcessedBlock ?? 0);
+    blockHeight.labels(chainId.toString()).set(lastProcessedBlock?.number ?? 0);
 
     // Start watching from (not including) the last processed block (if any)
     let fromBlock = lastProcessedBlock
-      ? lastProcessedBlock + 1
+      ? lastProcessedBlock.number + 1
       : this.deploymentBlock;
-    let currentBlockNumber = await provider.getBlockNumber();
+    let currentBlock = await provider.getBlock("latest");
 
     let printSyncInfo = true; // Print sync info only once
     let plan: ReplayPlan = {};
@@ -143,13 +154,14 @@ export class ChainContext {
     do {
       do {
         toBlock = !pageSize ? "latest" : fromBlock + (pageSize - 1);
-        if (typeof toBlock === "number" && toBlock > currentBlockNumber) {
-          // refresh the current block number
-          currentBlockNumber = await provider.getBlockNumber();
-          toBlock = toBlock > currentBlockNumber ? currentBlockNumber : toBlock;
+        if (typeof toBlock === "number" && toBlock > currentBlock.number) {
+          // refresh the current block
+          currentBlock = await provider.getBlock("latest");
+          toBlock =
+            toBlock > currentBlock.number ? currentBlock.number : toBlock;
 
           log.debug(
-            `Reaching tip of chain, current block number: ${currentBlockNumber}`
+            `Reaching tip of chain, current block number: ${currentBlock.number}`
           );
         }
 
@@ -185,24 +197,24 @@ export class ChainContext {
         if (typeof toBlock === "number") {
           fromBlock = toBlock + 1;
         }
-      } while (toBlock !== "latest" && toBlock !== currentBlockNumber);
-
-      const block = await provider.getBlock(currentBlockNumber);
+      } while (toBlock !== "latest" && toBlock !== currentBlock.number);
 
       // Replay only the blocks that had some events.
       for (const [blockNumber, events] of Object.entries(plan)) {
         log.debug(`Processing block ${blockNumber}`);
+        const historicalBlock = await provider.getBlock(Number(blockNumber));
         try {
           await processBlock(
             this,
-            Number(blockNumber),
+            historicalBlock,
             events,
-            block.number,
-            block.timestamp
+            currentBlock.number,
+            currentBlock.timestamp
           );
 
           // Set the last processed block to this iteration's block number
-          this.registry.lastProcessedBlock = Number(blockNumber);
+          this.registry.lastProcessedBlock =
+            blockToRegistryBlock(historicalBlock);
           await this.registry.write();
 
           // Set the block height metric
@@ -215,21 +227,21 @@ export class ChainContext {
       }
 
       // Set the last processed block to the current block number
-      this.registry.lastProcessedBlock = currentBlockNumber;
+      this.registry.lastProcessedBlock = blockToRegistryBlock(currentBlock);
 
       // Save the registry
       await this.registry.write();
 
       // It may have taken some time to process the blocks, so refresh the current block number
       // and check if we are in sync
-      currentBlockNumber = await provider.getBlockNumber();
+      currentBlock = await provider.getBlock("latest");
 
       // If we are in sync, let it be known
-      if (currentBlockNumber === this.registry.lastProcessedBlock) {
-        this.sync = ChainSync.SYNCED;
+      if (currentBlock.number === this.registry.lastProcessedBlock.number) {
+        this.sync = ChainSync.IN_SYNC;
       } else {
         // Otherwise, we need to keep processing blocks
-        fromBlock = this.registry.lastProcessedBlock + 1;
+        fromBlock = this.registry.lastProcessedBlock.number + 1;
         plan = {};
       }
     } while (this.sync === ChainSync.SYNCING);
@@ -247,7 +259,7 @@ export class ChainContext {
     }
 
     // Otherwise, run the block watcher
-    return await this.runBlockWatcher(watchdogTimeout);
+    return await this.runBlockWatcher(watchdogTimeout, currentBlock);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/utils/api.ts
```diff
@@ -33,6 +34,22 @@ export class ApiService {
...
+      const healths = contexts.map((context) => context.health);
+      const healthStatusDict = healths.reduce((acc, health) => {
+        acc[health.chainId] = health;
+        return acc;
+      }, {} as { [chainId: string]: ChainHealth });
+
+      const overallHealth = healths.every((health) => health.isHealthy);
+
+      res
+        .status(overallHealth ? 200 : 500)
```
alfetopito, line 50: I'm not sure 500 is the best for this situation.
After all, the service is able to handle the request and prepare a response without erroing internally.
It's just that some parts of the system aren't healthy.

Eliciting help from SO, there are the 2 interpretations in the top 2 answers: https://stackoverflow.com/questions/25389261/which-http-status-code-should-i-use-for-a-health-check-failure

Since additional info is sent as part of the payload, I'd say this should be used to determine the service status and if the service is able to respond successfully, it should be a 200 status code.

alfetopito: Got it.
Let's go with 5xx then.
What about 503, as pointed out in [this answer](https://stackoverflow.com/a/48005358)?

mfw78: Sounds great!

src/domain/chainContext.ts
```diff
@@ -24,6 +24,30 @@ const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
 
 const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
 
+enum ChainSync {
+  SYNCED = "SYNCED",
+  SYNCING = "SYNCING",
```
anxolin, line 29: how do native people pronounce this? cause i always pronounce it like "Sinking" 🚢💦

src/domain/chainContext.ts
```diff
@@ -24,6 +24,30 @@ const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
 
 const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
 
+enum ChainSync {
+  SYNCED = "SYNCED",
```
anxolin, line 28: ignored this, but `IN_SYNC` might be more visually different

src/domain/chainContext.ts
```diff
@@ -294,6 +323,26 @@ export class ChainContext {
...
+
+  get health(): ChainHealth {
+    return {
+      ...this.status,
+      isHealthy: this.isHealthy(),
+    };
+  }
+
+  private isHealthy(): boolean {
+    return this.sync === ChainSync.SYNCED;
```
anxolin, line 344: I think this is for the readyness probe, and not for the liveness probe, right?

src/utils/api.ts
```diff
@@ -33,6 +34,22 @@ export class ApiService {
     this.app.get("/", (_req: Request, res: Response) => {
       res.send("🐮 Moooo!");
     });
+    this.app.get("/health", async (_req: Request, res: Response) => {
```
anxolin, line 37: should we expose both the  `/health/liveness` and `/health/readyness` https://www.baeldung.com/linux/kubernetes-livenessprobe-readinessprobe#:~:text=Basics,healthy%20to%20serve%20incoming%20traffic.

src/utils/api.ts
```diff
@@ -33,6 +34,22 @@ export class ApiService {
...
+      // Using an iterator, process all the chain contexts, storing the health
+      // in a map, and if any of the contexts are unhealthy, return false
+      const contexts = Object.values(ChainContext.chains) as ChainContext[];
+      const healths = contexts.map((context) => context.health);
```
anxolin, line 41: why do you need casting here?

src/utils/api.ts
```diff
@@ -33,6 +34,22 @@ export class ApiService {
     this.app.get("/", (_req: Request, res: Response) => {
       res.send("🐮 Moooo!");
     });
+    this.app.get("/health", async (_req: Request, res: Response) => {
+      // Using an iterator, process all the chain contexts, storing the health
+      // in a map, and if any of the contexts are unhealthy, return false
```
anxolin, line 39: maybe shorter imperative comment? 

`Check the overall health. It's healthy if every network is healthy`

src/utils/api.ts
```diff
@@ -33,6 +34,22 @@ export class ApiService {
...
+      const healthStatusDict = healths.reduce((acc, health) => {
+        acc[health.chainId] = health;
+        return acc;
+      }, {} as { [chainId: string]: ChainHealth });
+
+      const overallHealth = healths.every((health) => health.isHealthy);
+
+      res
+        .status(overallHealth ? 200 : 500)
+        .send({ overallHealth, ...healthStatusDict });
```
anxolin, line 51: The logic, looks good to me, but for `readyness`, I would implement `liveness` based on the same criteria of watch dog (and remove watch dog --> or allow to disable it by config)

src/utils/api.ts
mfw78: Sorry, I didn't really make it clear in the PR about my thoughts on the various probes.

The states of the application are:
1. WarmUp
2. Synced

It doesn't _serve_ any data, so there I don't think there is a need for a readiness probe (as there are no ingress' that are being notified to turn on/off request flow to the pod).

Liveness could be exposed, but also if we remove the checks from the watch dog, this makes Kubernetes a requirement (which prevents self-sovereignty). Likely will use a CLI option for `--enable-watchdog` or similar

**NOTE**: A by-product of this is that the liveness probe specification for the IAC repo will require context as to which chain is being configured so it can modify the liveness probes.

src/domain/chainContext.ts
mfw78: It indeed is pronounced like "sinking". English is indeed silly. :sweat_smile:

src/utils/api.ts
mfw78: Done.

src/utils/api.ts
mfw78: Great catch. I don't - and it's gone :+1:

src/utils/api.ts
mfw78: Done.

src/domain/chainContext.ts
mfw78: Visually this makes sense. Done.

src/domain/chainContext.ts
```diff
@@ -24,6 +24,30 @@ const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
 
 const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
 
+enum ChainSync {
+  SYNCED = "SYNCED",
+  SYNCING = "SYNCING",
+}
```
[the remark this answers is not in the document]
mfw78: I was thinking about a third state similar to this, but I wasn't too fussed on `OUT_OF_SYNC` (how to handle times where there may be a very large gap in block production, longer than expected, but it actually _still_ is in sync). Instead I put in `UNKNOWN`.

src/domain/chainContext.ts
```diff
@@ -294,6 +323,26 @@ export class ChainContext {
       }
     }
   }
+
+  get status(): ChainStatus {
```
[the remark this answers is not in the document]
mfw78: OK, I've gone through and actually made it so that it will self-detect if it's running within a kubernetes environment. If it detects this, the watch-dog won't exit, but will instead set the chain sync to `UNKNOWN` status, leading the `/health` API endpoint to error with `503`, which can be picked up by directing the liveliness probe to `/health`.

src/domain/chainContext.ts
mfw78: > The kubelet uses readiness probes to know when a container is ready to start accepting traffic. A Pod is considered ready when all of its containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.

On this basis, I think we can actually use the same API endpoint (`/health`) to do both the startup probe and the liveness probe. As readiness probes are related to use of the pod within a service definition (to be load balanced again), this isn't a case that we're using, so I don't think we would implement a readiness probe?

src/utils/api.ts
mfw78: I think we can just use `/health` for both the startup probe and the liveness probe. No need for a ready probe as it's not being used in this context.

src/utils/api.ts
```diff
@@ -33,6 +34,22 @@ export class ApiService {
...
+      // Using an iterator, process all the chain contexts, storing the health
+      // in a map, and if any of the contexts are unhealthy, return false
+      const contexts = Object.values(ChainContext.chains) as ChainContext[];
+      const healths = contexts.map((context) => context.health);
+      const healthStatusDict = healths.reduce((acc, health) => {
+        acc[health.chainId] = health;
+        return acc;
+      }, {} as { [chainId: string]: ChainHealth });
+
+      const overallHealth = healths.every((health) => health.isHealthy);
```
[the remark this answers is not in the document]
mfw78: Absolutely, my bad. This has been refactored and moved into the `ChainContext` class.

src/utils/api.ts
mfw78: Readiness most probably isn't appropriate (see above comments). Liveness does use the same criteria as the watch dog, with the watch now setting sync status to `UNKNOWN` if it detects the timeout (and it's _not_ running in a kubernetes environment).

alfetopito (approved):

src/domain/chainContext.ts
```diff
@@ -24,6 +31,34 @@ const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
 
 const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
 
+enum ChainSync {
+  /** The chain is currently in the warm-up phase, synchronising from contract genesis or lastBlockProcessed */
+  SYNCING = "SYNCING",
+  /** The chain is in sync with the latest block */
+  IN_SYNC = "IN_SYNC",
+  /** The chain is in an unknown state based on duration of time since a block was processed */
+  UNKNOWN = "UNKNOWN",
```
anxolin, line 40: If this models that the last block you sold is old. Why not calling it `OUT_OF_SYNC` or `STALE` or similar?nit:

src/utils/context.ts
```diff
@@ -113,3 +113,11 @@ export function sendSlack(message: string): boolean {
   });
   return true;
 }
+
+export function isRunningInKubernetesPod(): boolean {
+  // Check if the standard Kubernetes environment variables are set
+  return (
+    process.env.KUBERNETES_SERVICE_HOST !== undefined &&
+    process.env.KUBERNETES_SERVICE_PORT !== undefined
```
anxolin, line 121: where are these in the docs/readme?

src/domain/chainContext.ts
```diff
@@ -276,48 +320,93 @@ export class ChainContext {
...
-      // If we haven't received a block in 30 seconds, exit so that the process manager can restart us
+      // If we haven't received a block within `watchdogTimeout` seconds, either signal
+      // an error or exit if not running in a kubernetes pod
       if (timeElapsed >= watchdogTimeout * 1000) {
+        const formattedElapsedTime = Math.floor(timeElapsed / 1000);
         log.error(
-          `Watchdog timeout (RPC failed, or chain is stuck / not issuing blocks)`
+          `Chain watcher last processed a block ${formattedElapsedTime}s ago (${watchdogTimeout}s timeout configured). Check the RPC.`
         );
+        if (isRunningInKubernetesPod()) {
```
anxolin, line 341: Why mentioning here Kubernetes? 
I think this could just be a setting: `exit on stale chain state`

```ts
this.sync = ChainSync.STALE;

if (exitOnStaleChainState) {
  await registry.storage.close();
  exit(1)
}
```

src/utils/api.ts
anxolin: What do you mean? In principle both are useful to kubernetes, right? 

so why not to separate and define them? 

What could be the discussion is, if readyness is just for when the API is up, or when the app is in SYNC. Maybe the first is just enough for the test

mfw78: > The kubelet uses readiness probes to know when a container is ready to start accepting traffic. A Pod is considered ready when all of its containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.

Readiness is not relevant to the watch-tower as the watch-tower isn't serving data for a kubernetes service, other than the metrics that are being scraped by prometheus. This probe is only useful when you are running pods that are providing services that are being load balanced (like websites, API services etc) and the pod is _temporarily unable to receive / process traffic_ - and signals such to a load balancer. As there is no redundancy running in the replica set, and therefore not using any fancy load balancing (doesn't fit our use case either to start with), there's no need for a discrete `readiness` probe.

This leaves us with `startupProbe` and `livelinessProbe`. The `livelinessProbe` will _not_ run before the `startupProbe` has succeeded. Therefore, as these can't run in parallel, we can actually just use the same API endpoint to provide for these (the configuration of the probes exists in the `Deployment` definition, and there's nothing prohibiting us from using the same endpoint for both).

src/utils/context.ts
mfw78: These are [automatically passed in by Kubernetes](https://kubernetes.io/docs/concepts/containers/container-environment/) when it starts the container (no configuration required on behalf of the watch-tower).

src/domain/chainContext.ts
mfw78: `STALE` sounds good. Done.

src/utils/api.ts
anxolin: we discussed in person. All good

PR #100, opened 2023-10-10
chore: rename some metrics

anxolin:
# Description
I just reviewed the names of the metrics again and they look great. With these ones, we have more than enough. Great job.

This PR suggests changing some naming on only a few of them

### DIFF
--- a/src/utils/metrics.ts
+++ b/src/utils/metrics.ts
@@ -69,8 +69,8 @@ export const reorgsTotal = new client.Counter({
   labelNames: ["chain_id"],
 });
 
-export const contractsTotal = new client.Counter({
-  name: "watch_tower_contracts_total",
+export const ownersTotal = new client.Counter({
+  name: "conditional_order_owners_total",
   help: "Total number of owner contracts",
   labelNames: ["chain_id"],
 });
--- a/src/utils/metrics.ts
+++ b/src/utils/metrics.ts
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
 });
 
 export const reorgDepth = new client.Gauge({
-  name: "watch_tower_reorg_depth",
-  help: "Depth of the current reorg",
+  name: "watch_tower_latest_reorg_depth",
+  help: "Depth of the last reorg",
   labelNames: ["chain_id"],
 });
 
-export const blockTime = new client.Gauge({
-  name: "watch_tower_block_time_seconds",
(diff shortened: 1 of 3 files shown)


src/utils/metrics.ts
```diff
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
 });
 
 export const reorgDepth = new client.Gauge({
-  name: "watch_tower_reorg_depth",
-  help: "Depth of the current reorg",
+  name: "watch_tower_latest_reorg_depth",
```
anxolin, line 49: if find this more clear

src/utils/metrics.ts
```diff
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
...
+  help: "Depth of the last reorg",
   labelNames: ["chain_id"],
 });
 
-export const blockTime = new client.Gauge({
-  name: "watch_tower_block_time_seconds",
+export const blockProducingRate = new client.Gauge({
```
anxolin, line 54: block time i would expect the latest epoc there, with this i try to show that it tracks the producing rate

src/utils/metrics.ts
```diff
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
...
+  name: "watch_tower_block_producing_rate_seconds",
   help: "Time since last block",
   labelNames: ["chain_id"],
 });
 
 export const eventsProcessedTotal = new client.Counter({
-  name: "watch_tower_events_processed_total",
-  help: "Total number of events processed",
+  name: "watch_tower_order_creation_events_total",
```
anxolin, line 61: events might be too generic. Its good to specify imo

src/utils/metrics.ts
```diff
@@ -69,9 +69,9 @@ export const reorgsTotal = new client.Counter({
   labelNames: ["chain_id"],
 });
 
-export const newContractsTotal = new client.Counter({
-  name: "watch_tower_new_contracts_total",
-  help: "Total number of new contracts",
+export const contractsTotal = new client.Counter({
```
anxolin, line 72: i don't love the previous nor the new suggestion. 
I feel the `contractsTotal` is also a bit generic. 

maybe `conditional_order_owners_total` is nicer?

anxolin: Actually i will change this in the PR (my first suggestion was `watch_tower_contracts_total`)

PR #102, opened 2023-10-10
Nits naming metrics2

anxolin:
# Description
Reapply https://github.com/cowprotocol/watch-tower/pull/100 but using branch `release/1.1.0` instead of `fix-memory-leak` branch

### DIFF
--- a/src/utils/metrics.ts
+++ b/src/utils/metrics.ts
@@ -69,8 +69,8 @@ export const reorgsTotal = new client.Counter({
   labelNames: ["chain_id"],
 });
 
-export const contractsTotal = new client.Counter({
-  name: "watch_tower_contracts_total",
+export const ownersTotal = new client.Counter({
+  name: "conditional_order_owners_total",
   help: "Total number of owner contracts",
   labelNames: ["chain_id"],
 });
--- a/src/utils/metrics.ts
+++ b/src/utils/metrics.ts
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
 });
 
 export const reorgDepth = new client.Gauge({
-  name: "watch_tower_reorg_depth",
-  help: "Depth of the current reorg",
+  name: "watch_tower_latest_reorg_depth",
+  help: "Depth of the last reorg",
   labelNames: ["chain_id"],
 });
 
-export const blockTime = new client.Gauge({
-  name: "watch_tower_block_time_seconds",
+export const blockProducingRate = new client.Gauge({
+  name: "watch_tower_block_producing_rate_seconds",
   help: "Time since last block",
   labelNames: ["chain_id"],
 });
 
 export const eventsProcessedTotal = new client.Counter({
-  name: "watch_tower_events_processed_total",
-  help: "Total number of events processed",
+  name: "watch_tower_order_creation_events_total",
+  help: "Total number of conditional order creation events processed",
   labelNames: ["chain_id"],
 });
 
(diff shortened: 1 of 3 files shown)


anxolin:
@mfw78 I will review the comments and address any concerns, but I would like to merge this to close the release (it will be too late for you to see this today, and i want to have this in staging asap)

Happy to revert any of this later if needed (for v1.2.0)

src/utils/metrics.ts
```diff
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
...
   labelNames: ["chain_id"],
 });
 
 export const eventsProcessedTotal = new client.Counter({
-  name: "watch_tower_events_processed_total",
-  help: "Total number of events processed",
+  name: "watch_tower_order_creation_events_total",
+  help: "Total number of conditional order creation events processed",
   labelNames: ["chain_id"],
 });
```
mfw78, line 64: I'm not sure about this as with the case of `MerkleRoot`, when doing `setRoot`, it's undefined as to whether you are actually _creating_ more orders or not (as the size of the merkle tree may actually be smaller, ie. deleting/removing orders).

src/utils/metrics.ts
```diff
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
...
 
-export const blockTime = new client.Gauge({
-  name: "watch_tower_block_time_seconds",
+export const blockProducingRate = new client.Gauge({
+  name: "watch_tower_block_producing_rate_seconds",
   help: "Time since last block",
   labelNames: ["chain_id"],
 });
```
mfw78, line 58: I'm unsure about this being labelled as a `rate`, as the measurement being taken is `scalar`. If we did want to stick to this kind of naming, it would be `watch_tower_block_production_rate_seconds`.

Even just writing this, usually when talking about a rate it's `units / sec`, where actually it could then be argued that in the case of mainnet, this metric would be expected to report 1 / 12. Per the [naming conventions](https://prometheus.io/docs/practices/naming/), with reference to `Power`, they make a statement about using a `rate()` function on a `joules` counter.

src/utils/metrics.ts
```diff
@@ -46,20 +46,20 @@ export const blockHeight = new client.Gauge({
...
 
 export const reorgDepth = new client.Gauge({
-  name: "watch_tower_reorg_depth",
-  help: "Depth of the current reorg",
+  name: "watch_tower_latest_reorg_depth",
+  help: "Depth of the last reorg",
```
mfw78, line 50: Suggest from a grammar side that the `help` uses the same tense, or we use something different completely such as:

`help: "Depth of the most recent reorg"`

mfw78 (commented):
Some response to nits

src/utils/metrics.ts
anxolin: @mfw78 good point, happy to change it to anything. I think just `events` its also to broad, there's for example https://github.com/cowprotocol/composable-cow/blob/main/src/ComposableCoW.sol#L46 as a event. 

What about `watch_tower_order_defintion_events_total`?

src/utils/metrics.ts
anxolin: What about just `watch_tower_block_production_seconds`?

src/utils/metrics.ts
anxolin: https://github.com/cowprotocol/watch-tower/pull/111

src/utils/metrics.ts
anxolin: https://github.com/cowprotocol/watch-tower/pull/111

src/utils/metrics.ts
anxolin: https://github.com/cowprotocol/watch-tower/pull/111

PR #106, opened 2023-10-11
feat: add orderbook API base url

anxolin:
# Description
Allow to set your own Orderbook API URL instead of the default ones from the SDK.

This PR is a follow up on https://github.com/cowprotocol/cow-sdk/pull/177.

The motivation, is to allow to set our service IP instead of the external one, to make the connection more performant.

## Bonus
While testing this, i realized that in case of an error, the SDK re-attempts 10 times using exponential backoff. This makes it take too much time in case of an error. I reduced to 5 the attempts, and now it re-attempts, but it will take at most a few seconds instead of 2-3 minutes.

## Follow up
After this is merged, we will need to set it up (please, let me do it!) 

See
https://cowservices.slack.com/archives/C05RMJB7ZFA/p1697021044757619?thread_ts=1697007472.259279&cid=C05RMJB7ZFA

## Test
Pass the `--orderBookApi ` and make sure it uses the parameter

```bash
ts-node ./src/index.ts run --rpc <your-rpc> --dployment-block 9847564 --page-size 5000 --orderBookApi https://api.cow.fi/goerli
```

Once this is in staging we can test this in the infrastructure project.

### DIFF
--- a/src/domain/chainContext.ts
+++ b/src/domain/chainContext.ts
@@ -36,6 +36,8 @@ const WATCHDOG_FREQUENCY = 5 * 1000; // 5 seconds
 
 const MULTICALL3 = "0xcA11bde05977b3631167028862bE2a173976CA11";
 
+export const SDK_BACKOFF_NUM_OF_ATTEMPTS = 5;
+
 enum ChainSync {
   /** The chain is currently in the warm-up phase, synchronising from contract genesis or lastBlockProcessed */
   SYNCING = "SYNCING",
@@ -107,6 +109,9 @@ export class ChainContext {
     this.orderBook = new OrderBookApi({
       chainId: this.chainId,
       baseUrls: this.orderBookApiBaseUrls,
+      backoffOpts: {
+        numOfAttempts: SDK_BACKOFF_NUM_OF_ATTEMPTS,
+      },
     });
 
     this.contract = composableCowContract(this.provider, this.chainId);
--- a/src/domain/chainContext.ts
+++ b/src/domain/chainContext.ts
@@ -9,7 +9,11 @@ import {
   RegistryBlock,
   blockToRegistryBlock,
 } from "../types";
-import { SupportedChainId, OrderBookApi } from "@cowprotocol/cow-sdk";
+import {
+  SupportedChainId,
+  OrderBookApi,
+  ApiBaseUrls,
+} from "@cowprotocol/cow-sdk";
 import { addContract } from "./addContract";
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -100,6 +100,16 @@ const deploymentBlockOption = new Option(
...
 
+const multiOrderBookApiOption = new Option(
+  "--orderBookApi <orderBookApi...>",
+  "Orderbook API base URLs (i.e. https://api.cow.fi/mainnet, https://api.cow.fi/xdai, etc.)"
+).default([]);
+
+const orderBookApiOption = new Option(
+  "--orderBookApi <orderBookApi>",
+  "Orderbook API base URL (i.e. https://api.cow.fi/mainnet)"
+).default(undefined);
```
mfw78, line 111: As this option doesn't have "make mandatory", is setting the default to `undefined` required?

src/index.ts
```diff
@@ -119,7 +130,8 @@ async function main() {
     .addOption(disableNotificationsOption)
     .addOption(slackWebhookOption)
     .action((options) => {
-      const { logLevel } = options;
+      const { logLevel, orderBookApi } = options;
+
```
mfw78, line 134: As the `orderBookApi` isn't being modified here, or checked for any sanity, is it required to be deconstructed here? It should just pass in with the `...options` below, no?

src/domain/chainContext.ts
```diff
@@ -120,7 +140,13 @@ export class ChainContext {
...
 
     // Save the context to the static map to be used by the API
-    const context = new ChainContext(options, provider, chainId, registry);
+    const context = new ChainContext(
+      options,
+      provider,
+      chainId,
+      registry,
+      orderBookApi
+    );
```
mfw78, line 149: As `orderBookApi` is part of the `options`, would we not just use the `options` within the constructor as this is basically already passed the whole way down and we can DRY this?

mfw78 (approved):
Great that we can configure this now. I think there's some scope for some minor DRY'ing.

src/index.ts
anxolin: 👍

src/index.ts
anxolin: https://github.com/cowprotocol/watch-tower/pull/106#event-10687336475

src/index.ts
anxolin: https://github.com/cowprotocol/watch-tower/pull/115

PR #110, opened 2023-10-12
feat: single chain

### LINKED ISSUE #71: feat: single-chain mode
### Problem
The current default mode for running the watch-tower is only multi-chain. This is a requirement for self-hosted (eg. dappnode), however there's minimal blast radius protection (one chain may affect another's watch-tower).

### Suggested solution
Implement a single-chain mode.

### Alternatives considered
Build blast radius protection into the current `run` mode of the watch-tower. This is likely an over-engineered solution that can be mitigated better by strict division between chain micro-services.

### Acceptance criteria

- [ ] `run` command moved to `run-multi`.
- [ ] `run` command implements a single chain's watch-tower.
- [ ] Command line configuration options for single-run are backwards compatible.

mfw78:
# Description
A long requested feature has been to enable single-chain running of the watch-tower (as is currently done on CoW Protocol infrastructure).

# Changes

- [x] Extracted all common options from `commander` to variables to DRY commander config.
- [x] Changed `run` to not take variadic `rpc` and `deploymentBlock`. Singular input only.
- [x] Introduced `run-multi` command that has the same functionality that `run` had previously.
- [x] Set all options to be configurable via environment variable, except variadic options on `run-multi`.
 
## How to test

1. Attempt to start a single chain as per instructions in the readme.

## Related Issues

Fixes #71

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -1,4 +1,5 @@
 export * from "./run";
+export * from "./runMulti";
 export * from "./replayBlock";
 export * from "./replayTx";
 export * from "./dumpDb";
--- a/src/commands/runMulti.ts
+++ b/src/commands/runMulti.ts
@@ -1,4 +1,4 @@
-import { RunSingleOptions } from "../types";
+import { RunMultiOptions } from "../types";
 import { getLogger, DBService } from "../utils";
 import { ChainContext } from "../domain";
 import { ApiService } from "../utils/api";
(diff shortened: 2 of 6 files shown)


anxolin (approved):
Great changes! now it looks easier to use.

One small improvement to make it dead simple, is to have a default value for the block number with the deployment block of each chain. This way, only one RPC is required

anxolin (approved):
Great changes! now it looks easier to use.

One small improvement to make it dead simple, is to have a default value for the block number with the deployment block of each chain. This way, only one RPC is required

Issue #72, opened 2023-09-30
feat: dappnode package


mfw78:
### Problem
For true self-sovereignty, and trustless environments, I should be able to easily run my own watch-tower. 

### Suggested solution
Provide a dappnode package that allows for monitoring one's own set of addresses across specifiable chains.

### Alternatives considered
Alternatives would be for users to run directly via CLI, or via docker (both reasonable options). Providing a dappnode package "neatens" this solution, and makes it more accessible.

### Acceptance criteria

- [x] Revise `run-multi` configuration to take a mix of RPC and deployment block (ie. `chain-config` is variadic and remove `deploymentBlock` option.
- [x] `address` option to filter monitor by a user specified addresses.
- [x] Entrypoint script for automatic configuration as much as possible.
- [x] Dappnode package defined with CI/CD to push to Pinata.

PR #112, opened 2023-10-15
chore: address filter and chain config

mfw78:
# Description
This PR tweaks some configuration options for easing deployment on dappnode.

# Changes

- [x] `rpc`, `deploymentBlock`, `watchdogTimeout`, and `orderBookApi` collapsed down into `chain-config` 
- [x] Added `ChainConfig` option parser
- [x] Added user-defined `address` filter (only process conditional orders from specific addresses)
- [x] Fixes a calculation issue with the watchdog
 
## How to test

1. Follow the readme instructions for quickly running
2. Observe no errors

## Related Issues

Fixes: #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,15 +33,22 @@ As an example, to run the latest version of the watch-tower via `docker`:
 docker run --rm -it \
   ghcr.io/cowprotocol/watch-tower:latest \
   run \
-  --chain-config <rpc>,<deployment-block>,<watchdog-timeout> \
+  --chain-config <rpc>,<deployment-block> \
   --page-size 5000
 ```
 
-**NOTE**: `watchdog-timeout` above is _optional_. If not specified, the default is 30 seconds. This is the timeout from which if the watch-tower does not receive a new block, it will exit, or signal non-live to Kubernetes. This is used to restart the watch-tower in the event of a RPC / network failure.
+**NOTE**: There are multiple optional arguments on the `--chain-config` parameter. For a full explanation of the optional arguments, use the `--help` flag:
 
-### Dappnode
+  ```bash
+  docker run --rm -it \
+    ghcr.io/cowprotocol/watch-tower:latest \
+    run \
+    --help
+  ```
 
-**TODO**: Add instructions for deploying to Dappnode.
+### DAppNode
+
+For [DAppNode](https://dappnode.com), the watch-tower is available as a package. This package is held in a [separate repository](https://github.com/cowprotocol/dappnodepackage-cow-watch-tower).
 
 ### Running locally
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ As an example, to run the latest version of the watch-tower via `docker`:
 docker run --rm -it \
   ghcr.io/cowprotocol/watch-tower:latest \
   run \
-  --chain-config <rpc-url>,<deployment-block>,<watchdog-timeout> \
+  --chain-config <rpc>,<deployment-block>,<watchdog-timeout> \
   --page-size 5000
 ```
 
@@ -56,7 +56,7 @@ docker run --rm -it \
 # Install dependencies
 yarn
 # Run watch-tower
-yarn cli run --chain-config <rpc-url>,<deployment-block> --page-size 5000
+yarn cli run --chain-config <rpc>,<deployment-block> --page-size 5000
 ```
 
 ## Architecture
@@ -179,7 +179,7 @@ It is recommended to test against the Goerli testnet. To run the watch-tower:
 # Install dependencies
 yarn
 # Run watch-tower
-yarn cli run --chain-config <rpc-url>,<deployment-block> --page-size 5000
+yarn cli run --chain-config <rpc>,<deployment-block> --page-size 5000
 ```
 
 ### Testing
--- a/README.md
+++ b/README.md
@@ -33,10 +33,11 @@ As an example, to run the latest version of the watch-tower via `docker`:
 docker run --rm -it \
   ghcr.io/cowprotocol/watch-tower:latest \
   run \
-  --rpc <rpc-url> \
-  --deployment-block <deployment-block> \
+  --chain-config <rpc-url>,<deployment-block>,<watchdog-timeout> \
   --page-size 5000
 ```
+
+**NOTE**: `watchdog-timeout` above is _optional_. If not specified, the default is 30 seconds. This is the timeout from which if the watch-tower does not receive a new block, it will exit, or signal non-live to Kubernetes. This is used to restart the watch-tower in the event of a RPC / network failure.
 
 ### Dappnode
 
@@ -55,7 +56,7 @@ docker run --rm -it \
 # Install dependencies
 yarn
 # Run watch-tower
-yarn cli run --rpc <rpc-url> --deployment-block <deployment-block> --page-size 5000
+yarn cli run --chain-config <rpc-url>,<deployment-block> --page-size 5000
 ```
 
 ## Architecture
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -284,6 +224,100 @@ function parseIntOption(option: string) {
...
+  if (!isValidUrl(rpc)) {
+    throw new InvalidArgumentError(
+      `${rpc} must be a valid URL (RPC) (chainConfig)`
+    );
+  }
+
+  const rawDeploymentBlock = parts[1];
+  // Ensure that the deployment block is a number
+  const deploymentBlock = Number(rawDeploymentBlock);
+  if (isNaN(deploymentBlock)) {
```
alfetopito, line 250: Nitpick: positive number?

src/index.ts
```diff
@@ -284,6 +224,100 @@ function parseIntOption(option: string) {
...
+
+  const rawWatchdogTimeout = parts[2];
+  // If there is a third part, it is the watchdogTimeout
+  const watchdogTimeout = parts.length > 2 ? Number(rawWatchdogTimeout) : 30;
+  // Ensure that the watchdogTimeout is a number
+  if (isNaN(watchdogTimeout)) {
+    throw new InvalidArgumentError(
+      `${rawWatchdogTimeout} must be a number (watchdogTimeout)`
+    );
+  }
```
alfetopito, line 264: Also might make sense the number is positive

src/types/index.ts
```diff
@@ -1,49 +1,56 @@
-export interface LogOptions {
+export type LogOptions = {
```
alfetopito, line 1: Go types!

README.md
```diff
@@ -33,11 +33,12 @@ As an example, to run the latest version of the watch-tower via `docker`:
...
   ghcr.io/cowprotocol/watch-tower:latest \
   run \
-  --rpc <rpc-url> \
-  --deployment-block <deployment-block> \
+  --chain-config <rpc>,<deployment-block>,<watchdog-timeout> \
   --page-size 5000
 ```
 
+**NOTE**: `watchdog-timeout` above is _optional_. If not specified, the default is 30 seconds. This is the timeout from which if the watch-tower does not receive a new block, it will exit, or signal non-live to Kubernetes. This is used to restart the watch-tower in the event of a RPC / network failure.
+
```
alfetopito, line 41: There's one more optional chain-config parameter not mentioned here in the read-me: orderbook api url

src/index.ts
mfw78: Great one. Done.

src/index.ts
mfw78: Likewise, done.

README.md
mfw78: In order to avoid just repeating the help written for commander, notes were added directing the user to `--help`.

src/types/index.ts
anxolin: Why these changes?

src/index.ts
```diff
@@ -63,65 +64,44 @@ const slackWebhookOption = new Option(
...
-).default([]);
+  .env("CHAIN_CONFIG")
+  .argParser(parseChainConfigOption);
 
-const orderBookApiOption = new Option(
-  "--orderBookApi <orderBookApi>",
-  "Orderbook API base URL (i.e. https://api.cow.fi/mainnet)"
-).default(undefined);
+const addressOption = new Option(
+  "--address <address...>",
```
anxolin, line 91: isn't this too generic name?

README.md
```diff
@@ -33,14 +33,22 @@ As an example, to run the latest version of the watch-tower via `docker`:
 docker run --rm -it \
   ghcr.io/cowprotocol/watch-tower:latest \
   run \
-  --rpc <rpc-url> \
-  --deployment-block <deployment-block> \
+  --chain-config <rpc>,<deployment-block> \
```
anxolin, line 36: I'm not sure this is a good change. It just couple the RPC and block

I still think the block should be independent and should have a default with the deployment block of the contracts. 

Then users can override. 

Although probably im missing the why this was necesarry, i guess has to do that in dappnode it was not simple to specify the params?

src/types/index.ts
anxolin: is this related in any sense to Dappnode?

src/domain/checkForAndPlaceOrder.ts
```diff
@@ -263,6 +263,8 @@ async function _processConditionalOrder(
         blockNumber,
       },
       provider,
+      // TODO: This should be DRY'ed. Upstream should take just an `orderBook` object that
```
anxolin, line 266: what do you mean `Upstream` here?

mfw78: `PollParams` should not take `orderbookApiConfig`, but should instead just take `orderBook` which is an instantiated Api with the parameters already set. Just like how the provider is already preset.

README.md
mfw78: In either which case, then the deployment block in that case would be optional in the `chain-config` tuple. Agreed that _eventually_ just publish a `composable-cow` NPM package that packages up all this info, that can then be re-exported by the `cow-sdk`. I don't think this is a high priority at the moment and would be better served when iterative work is done on the `composable-cow` repo with any new order types / interface revision. https://github.com/cowprotocol/composable-cow/issues/60 was created to specifically track this.

The reason to merge them together is to support multi-chain monitoring on DAppNode. Contrary to how the watch-tower is deployed on CoW infra in Kubernetes (where spinning up different deployments is trivial), writing a DAppNode package requires the exact number of containers to be specified for the `docker-compose`. Once the package is published, the number of containers is _not_ configurable / changeable. So one **cannot** dynamically select which networks they want to run, and have DAppNode add / remove containers from the `docker-compose` stack.

Alternatives to this include:

1. Expand the configuration options to allow for specific chains such as `--mainnet-rpc` etc.
2. Create a separate DAppNode package _per chain_.

Neither of the above alternatives are a maintainable solution. Both cause maintenance issues if new networks are added, and (2) incurs 3x the costs (each time a package is published to the DAppNode registry, it costs a transaction on Ethereum Mainnet).

src/index.ts
alfetopito: What about `safeAddress[es]`?

alfetopito (approved):

src/index.ts
anxolin: In other places we call it `owner`, not sure if its the best name

README.md
anxolin: > Expand the configuration options to allow for specific chains such as --mainnet-rpc etc.

This seems nicer, at least keeps the independence of the fields. Its possible that you will keep adding more fields of config per network, and i feel is a red flag to add them as comma separated list

what about, for multi-netork support we add:

First the list of networks:

```bash
# Either
---network 1 \
---network 2 \
---network 3 \

Or, a bit less nice
---network 1,2.3
```

Then the params per network
```bash
-- rpc-1 RPC1 \
-- rpc-2 RPC2 \
-- rpc-3 RPC3

# or using the name
-- rpc-mainnet RPC1 \
-- rpc-gc RPC2 \
-- rpc-goerli RPC3
```

This way, adding new params per network feels more natural.

mfw78: Having relationships between fields like this is a bit of a nightmare, and IMO is a great reason for a tuple. It's smelly, but I think there may be a more idiomatic way to do this. Isn't this essentially what is communicated by a URL? (ie. <protocol>://<domain>:<port>/<location>).

So, in which case, I think I may actually prefer to use a URL with optional query parameters to indicate fields that are optional. Something like:

```raw
<rpc>/<deploymentBlock>?watchdogTimeout=30&orderBookApiUrl=urlencodedField
```

For example, in DAppNode this would be:

```raw
http://erigon.dappnode:8545/17883049&watchdogTimeout=60&orderBookApiUrl=https%3A%2F%2Fapi.cow.fi%2Fmainnet
```

src/domain/checkForAndPlaceOrder.ts
mfw78: Tracked in https://github.com/cowprotocol/cow-sdk/issues/181

anxolin (approved):

src/index.ts
mfw78: I set this `--only-owner`

PR #116, opened 2023-11-19
feat: apply filtering policy

anxolin:
# Description
This PR implements a basic filtering policy for Programmable Orders.

## Configuration
You can define your own policy by defining your own file with the rules. 

The format allows you to SKIP or DROP orders who match some criteria:
```json
{
  "owners": {
    "0x0000000000000000000000000000000000000000": "SKIP"
  },
  "handlers": {
    "0x467751B9224f7828b37e2c242F7c37F1f8b91A0D": "DROP"
  }
}
```

## Initial config
The initial config lives now in:
https://github.com/cowprotocol/watch-tower/blob/config

There there's one file per network.

This branch is just temporal. The plan is to have the config done in https://github.com/cowprotocol/watch-tower-config which is a private repository, but it was remove from the scope of this PR.

## Filter Policy
In charge of the filtering. For now, I added only `pre-filters` but i could have postfilters in the future.
Also the rules are pretty simplistic, but future versions of the config could really have more advance filtering.

It can filter by `owner` and `handler`.

## Workflow
1. In every block, the policy is loaded before processing the conditional orders (TODO: maybe skip if there's none!)
2. For every conditional order processed, we apply the filters and depending on the result
* `ACCEPT`: Will let the order be processed (normal flow)
* `SKIP`: Will skip the order. Next block, we will decide again based on the configuration by the time is processed, if it's still `SKIP` will keep not processing the order.
* `DROP`:  Will skip the order for this run and any future blocks.

## Params for the CLI
The filtering is OPTIONAL, so by default no filtering will be done. 

EDIT: Removed the config param, now it uses the `--chain-config` (new way after the merge of https://github.com/cowprotocol/watch-tower/pull/112)
~You can pass optionally the URL of the configuration you want to use, for example:
`--filterPolicyConfig https://raw.githubusercontent.com/cowprotocol/watch-tower/config/filter-policy-100.json`~ 

You can specify the config as the 5th argument in ` --chain-config` argument

It also works for the multi-network run.

# Test

- Run without filtering
- Run with the initial config in `config` branch: `LOG_LEVEL=DEBUG,checkForAndPlaceOrder=DEBUG yarn ts-node ./src/index.ts run --rpc <rpc> --deployment-block 29389123 --page-size 5000 --filterPolicyConfig https://raw.githubusercontent.com/cowprotocol/watch-tower/config/filter-policy-100.json`
- Run with your own config, experiment with it

```
LOG_LEVEL=DEBUG,checkForAndPlaceOrder=DEBUG \
  yarn ts-node ./src/index.ts run \
  --chain-config <gc-rpc>,,,https://raw.githubusercontent.com/cowprotocol/watch-tower/config/filter-policy-100.json \
  --page-size 5000
```

# Out of scope
Add auth for the HTTPS request of the config file


src/domain/checkForAndPlaceOrder.ts
```diff
@@ -442,16 +469,26 @@ async function _placeOrder(params: {
...
+      receiver: order.receiver,
       feeAmount: order.feeAmount.toString(),
-      signingScheme: "eip1271",
+      validTo: order.validTo,
+      appData: order.appData,
+      partiallyFillable: order.partiallyFillable,
+      sellTokenBalance: order.sellTokenBalance,
+      buyTokenBalance: order.buyTokenBalance,
+      signingScheme: SigningScheme.EIP1271,
+      signature: order.signature,
```
anxolin, line 486: technically not part the PR. It was just annoying me in the logs. The spreading of `order` copy all props, including the `0`, `1`, `2` properties (its a pseudo-array)

With this, logs are less noisy and easier to read

shoom3301 (approved):

shoom3301:
Just approved, can't reason about these changes

src/utils/filterPolicy.ts
```diff
@@ -0,0 +1,84 @@
...
+   */
+  preFilter({ owner, conditionalOrderParams }: FilterParams): FilterAction {
+    if (!this.config) {
+      return FilterAction.ACCEPT;
+    }
+
+    const { owners, handlers } = this.config;
+
+    const action =
+      owners.get(owner) || handlers.get(conditionalOrderParams.handler);
```
mfw78, line 45: As a suggestion, are we sure we want to consider the individual owner case first? This may open up attacks if there's a public list with an owner explicitly whitelisted, in which case, one can create a bad handler and target that owner address with fake events being emitted.

mfw78 (commented):
Is there any reason why this is limited to being applied to `run-multi`, and hasn't been implemented for the `run` subcommand case?

mfw78 (approved):

src/utils/filterPolicy.ts
anxolin: For now don't make sense, as the default is to ACCEPT. It makes sense if we allow to change the default to DROP or SKIP (by the way, good feature and easy to add)

However, if we change it in the future, we might not remember this, so changing it just in case.

anxolin:
> Is there any reason why this is limited to being applied to run-multi, and hasn't been implemented for the run subcommand case?

The `run` don't require changes. It has already the param. If you try the command above should work

PR #125, opened 2023-12-27
feat: websockets rpc provider

mfw78:
# Description

This PR changes to use a websockets RPC provider

# Changes

- [x] Use websockets

## How to test
1. Run in staging!

### DIFF
--- a/src/domain/chainContext.ts
+++ b/src/domain/chainContext.ts
@@ -31,7 +31,7 @@ import {
   reorgDepth,
   reorgsTotal,
 } from "../utils/metrics";
-import { ConnectionInfo, hexZeroPad } from "ethers/lib/utils";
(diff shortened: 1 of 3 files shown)


anxolin (commented):
I'm not very happy with these changes. I feel a change such as this should be discussed.

It removes features, and doesn't have any reviews. I fear we are passing too many uncontrolled changes, without enough eyes or agreement.

In particular, given the requirements of the watch tower, a Http RPC provider is good. It's the responsibility of the one using it, to decide on how to solve RPC issues, maybe they want to use a local node which doesn't have load balancing and don't expose Websockets. 
If you want Websockets, you can recommend it in the README when you talk about config, but not force people. 

Besides kills the authentication logic, and makes the code less generic as, before we were getting the provider from a method, and now we instanciate an specific provider.

PR #126, opened 2024-01-10
chore: update filter policy hourly

mfw78:
# Description

Update the filter policy every hour

# Changes

- [x] Set frequency at which to update the filter policy

### DIFF
--- a/src/domain/chainContext.ts
+++ b/src/domain/chainContext.ts
@@ -469,8 +469,11 @@ async function processBlock(
     block.number.toString()
   );
 
-  // Get the latest filter policy
-  if (filterPolicy) {
+  // Refresh the policy every hour
+  // NOTE: This is a temporary solution until we have a better way to update the filter policy
+  const filterFrequency =
+    3600 / (context.chainId === SupportedChainId.GNOSIS_CHAIN ? 5 : 12); // 5 seconds for gnosis, 12 seconds for mainnet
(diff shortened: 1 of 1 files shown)


mfw78:
Merging for hot fix

src/domain/chainContext.ts
```diff
@@ -469,8 +470,15 @@ async function processBlock(
...
-  // Get the latest filter policy
-  if (filterPolicy) {
+  // Refresh the policy every hour
+  // NOTE: This is a temporary solution until we have a better way to update the filter policy
+  const blocksPerFilterFrequency =
+    FILTER_FREQUENCY_SECS /
+    (context.chainId === SupportedChainId.GNOSIS_CHAIN ? 5 : 12); // 5 seconds for gnosis, 12 seconds for mainnet
+  if (
+    filterPolicy &&
+    block.number % (FILTER_FREQUENCY_SECS / blocksPerFilterFrequency) == 0
```
anxolin, line 480: looks too complicated... 
why not just forget about the time frequency and define the block frequency
- 720 for gnosis 
- 300 for other networks

also, better to use constant for these numbers (no magic numbers like 5 or 12)

same effect, but easier to read.

anxolin (commented):
Thanks for the fix. A bit concerned this will make the policy be effective only after one hour

I think we could do it more often, right?

PR #131, opened 2024-01-17
Configuration file

mfw78:
# Description

Configuration methodology (via CLI options) has been outgrown with the number of options. This PR removes all chain / network related configuration and moves to a configuration file.

# Changes

- [x] Move all configuration for chains to a dedicated file
- [x] Remove unused features (replay tx / replay block)
- [x] Eliminates dynamic download of filter policies (now within configuration file)
- [x] Update `README.md`

## How to test
1. Adjust the `config.json.example` for suitable configuration
2. Run and verify sync process completes warmup

## Related Issues

Fixes #129

PR #135, opened 2024-01-19
chore: pr 131 comments

mfw78:
# Description

This PR contains fixes for comments from #131 

# Changes

- [x] Fix readme error
- [x] Cleanup unused type in `FilterPolicy`
- [x] Add gitignore for default config file

## How to test
1. Follow `README.md` instructions for running locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,4 @@ yalc.lock
 database*
 
 # Default config
-config.json
+/config*.json
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ yalc.lock
 
 # Databases
 database*
+
+# Default config
+config.json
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -17,4 +17,7 @@ node_modules/
 yalc.lock
 
 # Databases
-database*
\ No newline at end of file
+database*
+
+# Default config
+config.json
```
anxolin, line 23: sorry, it was me who suggested. but maybe we can make the pattern:

```
/config*.json
````

As we can have all sort of config files. 



image


Also, i made it relative to the root, so this don't byte us if we create a config somewhere in the `src`

anxolin (approved):
Approve with comments

.gitignore
mfw78: In general, I'd be a fan of only ignoring the recommended / assumed default configuration (which is also that which we document), however it's low hanging fruit to ignore all configs. This has been included :+1:

PR #141, opened 2024-01-23
refactor: modular architecture

mfw78:
# Description

This PR refactors legacy locations of code to more accurately reflect the area of responsibility for the respective code.

# Changes

- [x] Any core "service" that is running is located within the "services" module.
- [x] Domain logic is split related to indexing ("events"), and polling ("polling")
- [x] Filtering that has been applied is made more meaningful and moved to a "filtering" module

## How to test
1. Verify that `yarn test` passes
2. Run from genesis on `mainnet` and verify the block watcher comes online (passes `warmUp`).

### DIFF
--- a/src/commands/dumpDb.ts
+++ b/src/commands/dumpDb.ts
@@ -1,5 +1,6 @@
 import { DumpDbOptions, Registry } from "../types";
-import { DBService, getLogger } from "../utils";
+import { DBService } from "../services";
+import { getLogger } from "../utils";
 
 /**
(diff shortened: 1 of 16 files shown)


alfetopito (approved):

anxolin (approved):
Soft approve

I don't see the functional difference between `domain` and `services`.

Also, no need to address, but my 2c. If we are structuring in layers, i think i miss a bit the layer for DAO or Repository layer. 

So I would leave these 4 =>  commands, services, repository, model
Each layer interacts with the one below.

PR #145, opened 2024-02-14
Add conditional order id to our model

anxolin:
# Description
Ths PR adds the conditional order ID to our mode.

As this is a breaking change, the PR also does some model migration (from version 1 to version 2).

Additionally, it adds some additional logging to print the conditional order ID which is convenient.

### DIFF
--- a/src/domain/events/index.ts
+++ b/src/domain/events/index.ts
@@ -18,9 +18,9 @@ import {
   Proof,
   Registry,
 } from "../../types";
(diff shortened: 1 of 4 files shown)


mfw78 (approved):
Looks good! One to be aware of, in the future (or even possible with TWAP), the `conditionalOrderId` is _not_ unique and can result in interesting issues when searching via logs.

PR #147, opened 2024-02-14
fix: block watcher metrics and last processed block

mfw78:
# Description

Fixes an issue associated with block processing unexpected errors.

# Changes

- [x] Move metrics to `finally`
- [x] Move registry saving to `finally`

## How to test
1. Observe in staging

### DIFF
--- a/src/services/chain.ts
+++ b/src/services/chain.ts
@@ -325,8 +325,11 @@ export class ChainContext {
     log.debug(`Watchdog timeout: ${watchdogTimeout} seconds`);
     let lastBlockReceived = lastProcessedBlock;
     provider.on("block", async (blockNumber: number) => {
+      // Intentionally don't guard the getBlock call with a try/catch block
+      // because we want to crash the process if this fails. This will result
+      // in the kubernetes pod restarting and trying to recover.
+      const block = await provider.getBlock(blockNumber);
       try {
-        const block = await provider.getBlock(blockNumber);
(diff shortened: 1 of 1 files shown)


src/services/chain.ts
```diff
@@ -352,23 +355,15 @@ export class ChainContext {
...
-        log.error(
-          `Error in pollContractForEvents for block ${blockNumber}`,
-          error
-        );
+        log.error(`Error processing block ${blockNumber}`, error);
+      } finally {
+        // Block height metric
+        this.registry.lastProcessedBlock = blockToRegistryBlock(block);
+        await this.registry.write();
+        metrics.blockHeight.labels(chainId.toString()).set(blockNumber);
```
anxolin, line 366: but if the block was not process, why should we write it is. 
Wouldn't this cover up next issue we have?

src/services/chain.ts
```diff
@@ -325,8 +325,11 @@ export class ChainContext {
     log.debug(`Watchdog timeout: ${watchdogTimeout} seconds`);
     let lastBlockReceived = lastProcessedBlock;
     provider.on("block", async (blockNumber: number) => {
+      // Intentionally don't guard the getBlock call with a try/catch block
+      // because we want to crash the process if this fails. This will result
+      // in the kubernetes pod restarting and trying to recover.
```
anxolin, line 330: i would rephrase and not reference kubernetes here, as this is code that can be run in various ways

PR #151, opened 2024-05-10
Drop ComposableCoW bytecode check

mfw78:
# Description
Removes the janky bytecode check that was being used to enforce that the contract being indexed was `ComposableCoW` compatible. This is possible given that there are exhaustive switch statements that were published for the revert messages.

# Changes

- [x] Add  a try/catch around the decode function result to catch bad `ComposableCoW`-like implementations.
- [x] Remove bytecode checks.

## How to test
1. Remove the sepolia database.
2. Restart
3. Observe new CoW AMMs :rocket:

### DIFF
--- a/src/domain/polling/index.ts
+++ b/src/domain/polling/index.ts
@@ -665,11 +665,11 @@ async function _pollLegacy(
 ): Promise<PollResult> {
   const { contract, multicall, chainId } = context;
   const log = getLogger("checkForAndPlaceOrder:_pollLegacy", orderRef);
+  const { composableCow: target } = conditionalOrder;
   const { handler } = conditionalOrder.params;
   // as we going to use multicall, with `aggregate3Value`, there is no need to do any simulation as the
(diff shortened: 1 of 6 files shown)


src/domain/polling/index.ts
```diff
@@ -690,16 +690,27 @@ async function _pollLegacy(
...
+          order,
+          signature,
+        };
+      } catch (error: any) {
+        log.error(`ethers/decodeFunctionResult Unexpected error`, error);
+        metrics.pollingOnChainEthersErrorsTotal.labels(...metricLabels).inc();
+        return {
+          result: PollResultCode.DONT_TRY_AGAIN,
+          reason:
+            "UnexpectedErrorName: Unexpected error" +
```
fedgiac, line 710:
```suggestion
            "UnexpectedErrorName: Data decoding failure" +
```

fedgiac (approved):
Looks good!

PR #154, opened 2024-06-18
feat: improve first sync

anxolin:
# Description
This PR is an attempt to improve the first sync for watch-tower

## Context: Before this PR
The issue with the current model is that
1) It enter SYNC mode
2) In SYNC mode it will check all pending blocks from last processed until the tip of the blockchains
- Instead of actually processing the blocks, its just "saving" the block for processing later (its creating a sync plan)
3) Once in SYNC, it will apply the plan
- This time it will process for real the blocks

The issues we experience is becasue between 2 and 3 there's a delay and the metric of "blocks processed" goes down. Thats our main metric.

The solution is to not make a plan and process blocks as we find them instead of  waiting to be fully catched up.

## Proposed solution
I just refactored the 2 points where we do the processing of blocks (applying the plan, and when watching for new blocks). 

Now I use an auxiliary function that will process the block, update the metrics, and persist in the database. 

This should help in several ways:
- We will improve the metric of "blocks processed". This metric is key for our alerts. We don't want to be notified if during a restart watchtower needs some time to consume all pending blocks
- Additionally should help with very big syncs, like first syncs or big downtimes. Before this PR it was indexing all blocks before processing any block. This approach should be better because in case of a crash, next run will resume the work where it left it.
- Derived from the item above, I suspect the memory issues could be related to applying a big plan when the pod restarts

## Test
I haven't tested this much and is very sensitive change. I would love to get some feedback first, then I'd like to test it in staging.  

I just did a minimal test of running it in Arbitrum locally:

image

The watch-tower arrived to a SYNC state and processed blocks as it found them:
image

### DIFF
--- a/src/services/chain.ts
+++ b/src/services/chain.ts
@@ -254,25 +254,33 @@ export class ChainContext {
           });
         }
 
+        // Persist the last processed block (we are in sync up until toBlock)
+        persistLastProcessedBlock({
+          context: this,
+          block: await provider.getBlock(toBlock),
+          log,
+        });
+
         // only possible string value for toBlock is 'latest'
         if (typeof toBlock === "number") {
           fromBlock = toBlock + 1;
         }
       } while (toBlock !== "latest" && toBlock !== currentBlock.number);
 
-      // Set the last processed block to the current block number
-      this.registry.lastProcessedBlock = blockToRegistryBlock(currentBlock);
+      const lastProcessedBlock = this.registry.lastProcessedBlock?.number || 0;
+      // // Set the last processed block to the current block number
+      // this.registry.lastProcessedBlock = blockToRegistryBlock(currentBlock);
 
       // It may have taken some time to process the blocks, so refresh the current block number
       // and check if we are in sync
       currentBlock = await provider.getBlock("latest");
 
       // If we are in sync, let it be known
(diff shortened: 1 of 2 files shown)


src/services/chain.ts
```diff
@@ -228,67 +227,57 @@ export class ChainContext {
...
           }
 
-          plan[event.blockNumber].add(event);
+          return acc;
+        }, {});
+
+        // Process blocks in order
+        for (const blockNumberKey of Object.keys(eventsByBlock).sort()) {
+          const blockNumber = Number(blockNumberKey);
+          lastProcessedBlock = await processBlockAndPersist({
```
alfetopito, line 247: Here `lastProcessedBlock` is set.
However, I don't see it used anywhere before it's overwritten on line `258`.

alfetopito (approved):

src/services/chain.ts
anxolin: Yes, you are right. I deleted it 858c00fd8006557182d84f3824c17d6d22c5dbf5

I think is just because during the warmup we search for events for a block range, once we find an "interesting block" we update the DB and prometheus metrics to signal that that's the last processed block. So that variable will always have the latest indexed block (even if we don't use it)

I could add a log entry instead of just deleting, but anyways, I decided to kill the assinment.

mfw78 (approved):

PR #156, opened 2024-07-19
Throttle block consumption

anxolin:
# Description
This PR adds a new optional config parameter to each network, allowing us to define some block throttling. 

the setting is called `processEveryNumBlocks` : 
* Throttle block processing to only process blocks every N blocks. Set to 1 to process every block, 2 to process every other block, etc.

Also, note that even if we skip a block, it doesn't mean we don't have any work on it. We still need to check if there's new orders coming in. So what is actually skipped is only the programmatic order checks and posting orders. 

Also, it persists all Prometheus metrics, so it should count as a "consumed block" for our consumption rate alerts, meaning we should see the block consumption rate progressing faster. 

image

This is an example on how to use it: `"processEveryNumBlocks": 40,`
This would do polling and post orders once every 40 blocks. In arbitrum this should be every 8 seconds

```yaml
{
  "networks": [
    {      
      "name": "arbitrum-one",
      "rpc": "wss://your-rpc/arbitrum",
      "deploymentBlock": 204704802,
      "processEveryNumBlocks": 40,
      "filterPolicy": {
        "defaultAction": "DROP",
        "handlers": {
          "0x44569Cbd4E10dd5e97293337964Eff32d58ed352": "ACCEPT",
          "0x519BA24e959E33b3B6220CA98bd353d8c2D89920": "ACCEPT",
          "0x6cF1e9cA41f7611dEf408122793c358a3d11E5a5": "ACCEPT",
          "0xd3338f21c89745e46af56aeaf553cf96ba9bc66f": "ACCEPT",
          "0xE8212F30C28B4AAB467DF3725C14d6e89C2eB967": "ACCEPT",
          "0xB148F40fff05b5CE6B22752cf8E454B556f7a851": "ACCEPT"
        }
      },
      "watchdogTimeout": 120
    }
  ]
}
```

### DIFF
--- a/src/types/types.d.ts
+++ b/src/types/types.d.ts
@@ -13,6 +13,10 @@ export interface Config {
     rpc: string;
     deploymentBlock: number;
     watchdogTimeout?: number;
+    /**
+     * Throttle block processing to only process blocks every N blocks. Set to 1 to process every block, 2 to process every other block, etc.
+     */
+    processEveryNumBlocks?: number;
     orderBookApi?: string;
     pageSize?: number;
     filterPolicy: {
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22-alpine AS build
+FROM node:22.2-alpine AS build
 
 WORKDIR /usr/src/app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:alpine AS build
+FROM node:22-alpine AS build
 
 WORKDIR /usr/src/app
 
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "lint": "eslint \"src/**/*.ts\" && prettier --check \"src/**/*.ts\"",
     "lint:fix": "eslint --fix \"src/**/*.ts\" && prettier --write \"src/**/*.ts\"",
     "test": "yarn build && jest ./dist",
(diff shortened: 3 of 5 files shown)


src/types/types.d.ts
```diff
@@ -13,6 +13,10 @@ export interface Config {
     rpc: string;
     deploymentBlock: number;
     watchdogTimeout?: number;
+    /**
+     * Throttle block processing to only process blocks every N blocks. Set to 1 to process every block, 2 to process every other block, etc.
+     */
```
alfetopito, line 18: Can you mention the default behaviour?

alfetopito (approved):

src/types/types.d.ts
```diff
@@ -13,6 +13,10 @@ export interface Config {
     rpc: string;
     deploymentBlock: number;
     watchdogTimeout?: number;
+    /**
+     * Throttle block processing to only process blocks every N blocks. Set to 1 to process every block, 2 to process every other block, etc.
```
fleupold, line 17: nit: I feel this comment is a bit redundant (it's documented in the shema above and we don't have redundant docs for any of the other parameters)

fleupold (approved):
Send it!

src/types/types.d.ts
anxolin: This comment is automatically generated from the JSON schema

PR #159, opened 2024-10-03
fix: save orders more often

alfetopito:
# Description

This PR introduces more often saving of orders after processing.
For now, after every 20 orders processed, it'll trigger a db save.

There are also some logging improvements to make debugging easier.

## How to test

Run the service.
It should start and process orders as usual.

## Related Issues

There have been performance issues which I'm not able to tell what's causing it so far.
This change is an attempt to alleviate the in memory load by saving it more often.

### DIFF
--- a/src/domain/polling/index.ts
+++ b/src/domain/polling/index.ts
@@ -8,13 +8,6 @@ import {
 import { ethers } from "ethers";
 import { BytesLike } from "ethers/lib/utils";
 
-import { ConditionalOrder, OrderStatus } from "../../types";
-import {
-  formatStatus,
-  getLogger,
-  handleOnChainCustomError,
-  metrics,
-} from "../../utils";
 import {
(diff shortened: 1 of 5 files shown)


shoom3301 (approved):
LGTM.
Checked the logic, works as expected:

```
var items = new Set([1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20])
var itemsToDelete = []
var index = 0

for (const item of items) {
    index++

    if (index % 5 === 1 && index > 1) {
        for (const itemToDelete of itemsToDelete) {
          const deleted = items.delete(itemToDelete)
          console.log('delete', itemToDelete, deleted)
        }
        itemsToDelete = []
    }
    
    console.log('item', item)
    itemsToDelete.push(item)
}

console.log('items result', items)
```

image

PR #161, opened 2024-10-08
fix: attempt to fix memory leak

alfetopito:
# Description

I noticed that since my last changes #159 and #160 , there has been a very visible memory leak, specially on arb1.
image

[graph link](https://g-0263500beb.grafana-workspace.eu-central-1.amazonaws.com/goto/sjeXFwkHg?orgId=1)

While I have not been able to detect where the memory leak is, this PR has several improvements that might help.

# Changes

- [x] Remove a couple of unnecessary Array conversions, using iterators instead
- [x] Use a single logger instance in the Registry class
- [x] Move block skipping logic early, avoiding unnecessary RPC calls
- [x] Reuse `block` instance if already queried

## How to test
Run it locally: should work
image

~I'll also run it on staging for awhile before taking this PR out of draft.~ can't apply it, pulumi is not happy with me.

### DIFF
--- a/src/services/chain.ts
+++ b/src/services/chain.ts
@@ -320,7 +320,17 @@ export class ChainContext {
     let lastBlockReceived = lastProcessedBlock;
     provider.on("block", async (blockNumber: number) => {
       try {
+        // Decide if we should process this block before polling for events
+        const shouldProcessBlock =
+          blockNumber % this.processEveryNumBlocks === 0;
+
+        if (!shouldProcessBlock) {
+          log.debug(`Skipping block ${blockNumber}`);
+          return;
+        }
+
         log.debug(`New block ${blockNumber}`);
+
         const block = await provider.getBlock(blockNumber);
 
         // Set the block time metric
@@ -442,7 +452,7 @@ async function processBlock(
   blockNumberOverride?: number,
   blockTimestampOverride?: number
 ) {
-  const { provider, chainId, processEveryNumBlocks } = context;
+  const { provider, chainId } = context;
   const timer = metrics.processBlockDurationSeconds
     .labels(context.chainId.toString())
     .startTimer();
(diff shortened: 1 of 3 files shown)


shoom3301 (approved):

src/services/chain.ts
```diff
@@ -320,7 +320,17 @@ export class ChainContext {
     let lastBlockReceived = lastProcessedBlock;
     provider.on("block", async (blockNumber: number) => {
       try {
+        // Decide if we should process this block before polling for events
+        const shouldProcessBlock =
+          blockNumber % this.processEveryNumBlocks === 0;
+
+        if (!shouldProcessBlock) {
+          log.debug(`Skipping block ${blockNumber}`);
+          return;
```
anxolin, line 329: but we can't just skip the block without fetching the events for it...

We can skip the polling, but if we skip the whole block we will miss all orders created in the block 😅

src/services/chain.ts
```diff
@@ -563,27 +575,25 @@ async function pollContractForEvents(
...
-        return null;
+        });
       }
-    })
-    .filter((e): e is ConditionalOrderCreatedEvent => e !== null)
-    .filter((e): e is ConditionalOrderCreatedEvent => {
-      return addresses ? addresses.includes(e.args.owner) : true;
-    });
+    } catch {
+      // Ignore errors and do not add to the accumulator
```
anxolin, line 593: I know it was already like that, but why do we want to ignore the erros?

src/types/model.ts
```diff
@@ -189,7 +190,11 @@ export class Registry {
   }
 
   get numOrders(): number {
-    return Array.from(this.ownerOrders.values()).flatMap((o) => o).length;
+    let count = 0;
+    for (const orders of this.ownerOrders.values()) {
+      count += orders.size; // Count each set's size directly
+    }
+    return count;
```
anxolin, line 197: ubernit: a reduce here would have been more sutiable

PR #163, opened 2024-11-28
fix: fix missing orders

anxolin:
# Description

## Context
@elena-zh create a TWAP tx that didn't get index in watch tower.
https://basescan.org/tx/0xa9e88af5dee1d51f217bbe7613e2897cec6b3d989e8f3cab5142aa74ccb96278#eventlog 

I saw there was not logs for that transaction. Which is odd! I can see the event for conditional order created.
During my test, my transaction was indexed. 
https://basescan.org/tx/0xf0a848e6899330e84e86160f079cf341b145b1fb2bbc2f7a8aec1dfb425384c9#eventlog 

In principle both TX emit the same event, and should have been picked in the same way, but something was making us to skip some indexing of events

## Cause
I reviewed the code and saw some changes applied 2 months ago were making us skip the indexing of events for some blocks.

See:
https://github.com/cowprotocol/watch-tower/pull/161/files#r1862591918

## Solution
Bring back the code that skips the polling of tradable orders from the handlers, but leaves the indexing of conditional orders (so we don't miss an order creation and we can do the polling in a future block).

image

### DIFF
--- a/src/services/chain.ts
+++ b/src/services/chain.ts
@@ -320,15 +320,6 @@ export class ChainContext {
     let lastBlockReceived = lastProcessedBlock;
     provider.on("block", async (blockNumber: number) => {
       try {
-        // Decide if we should process this block before polling for events
-        const shouldProcessBlock =
-          blockNumber % this.processEveryNumBlocks === 0;
-
-        if (!shouldProcessBlock) {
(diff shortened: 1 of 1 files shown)


anxolin:
Merging asap to have it working in staging ASAP. Please @shoom3301 could you review

PR #164, opened 2024-12-02
chore: add more logs for OOM search

shoom3301:
This code doesn't change the logic, it only adds more logs and refactors the code a bit.

1. Added more logs
2. Removed excessive exports
3. Got rid of excessive ComposableCoW contract instances
4. Added "Conditional order params are equal but not the same" log

The last point is my suspect. I assume that comparation here is not valid, we should not use shallow comparison for the objects.

### DIFF
--- a/src/domain/events/index.ts
+++ b/src/domain/events/index.ts
@@ -10,7 +10,6 @@ import { BytesLike, ethers } from "ethers";
 import {
   ComposableCoW,
   ComposableCoW__factory,
-  ComposableCoWInterface,
   ConditionalOrderCreatedEvent,
   IConditionalOrder,
   MerkleRootSetEvent,
@@ -21,6 +20,8 @@ import {
 import { ConditionalOrder, ConditionalOrderParams } from "@cowprotocol/cow-sdk";
 
 import { ChainContext } from "../../services/chain";
+
+const composableCow = ComposableCoW__factory.createInterface();
 
 /**
  * Listens to these events on the `ComposableCoW` contract:
@@ -48,7 +49,6 @@ async function _addContract(
   context: ChainContext,
   event: ConditionalOrderCreatedEvent
 ) {
-  const composableCow = ComposableCoW__factory.createInterface();
   const log = getLogger("addContract:_addContract");
   const { registry } = context;
   const { transactionHash: tx, blockNumber } = event;
@@ -57,11 +57,7 @@ async function _addContract(
   let hasErrors = false;
   let numContractsAdded = 0;
 
-  const { error, added } = await registerNewOrder(
-    event,
-    composableCow,
-    registry
-  );
+  const { error, added } = await registerNewOrder(event, registry);
 
   if (added) {
     metrics.ownersTotal.labels(context.chainId.toString()).inc();
@@ -91,7 +87,6 @@ async function _addContract(
 
 async function registerNewOrder(
   event: ConditionalOrderCreatedEvent | MerkleRootSetEvent,
-  composableCow: ComposableCoWInterface,
   registry: Registry
 ): Promise<{ error: boolean; added: boolean }> {
   const log = getLogger("addContract:registerNewOrder");
--- a/src/domain/events/index.ts
+++ b/src/domain/events/index.ts
@@ -216,6 +216,7 @@ function add(
         const areConditionalOrderParamsEqual =
           getAreConditionalOrderParamsEqual(conditionalOrder.params, params);
 
+        // TODO: delete this log after testing
         if (
           areConditionalOrderParamsEqual &&
           conditionalOrder.params !== params
--- a/src/domain/events/index.ts
+++ b/src/domain/events/index.ts
@@ -22,9 +22,6 @@ import { ConditionalOrder, ConditionalOrderParams } from "@cowprotocol/cow-sdk";
 
 import { ChainContext } from "../../services/chain";
 
-const composableCow = ComposableCoW__factory.createInterface();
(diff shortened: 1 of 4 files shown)


shoom3301:
I have read the CLA Document and I hereby sign the CLA

src/domain/events/index.ts
```diff
@@ -21,6 +22,9 @@ import { ConditionalOrder, ConditionalOrderParams } from "@cowprotocol/cow-sdk";
 
 import { ChainContext } from "../../services/chain";
 
+const composableCow = ComposableCoW__factory.createInterface();
+const log = getLogger("addContract:_addContract");
```
anxolin, line 26: Why is this log saying `addContract` if its just on top of this file which is called `events`?

src/domain/events/index.ts
```diff
@@ -86,15 +90,16 @@ async function _addContract(
   }
 }
 
-export async function _registerNewOrder(
+async function registerNewOrder(
   event: ConditionalOrderCreatedEvent | MerkleRootSetEvent,
   composableCow: ComposableCoWInterface,
```
anxolin, line 95: Why do yo need to receive the `composableCow` if you already have it in scope?

src/domain/events/index.ts
```diff
@@ -206,6 +212,23 @@ export function add(
     // Iterate over the conditionalOrders to make sure that the params are not already in the registry
     for (const conditionalOrder of conditionalOrders?.values() ?? []) {
       // Check if the params are in the conditionalOrder
+      if (conditionalOrder) {
```
anxolin, line 215: maybe add a TODO to delete these lines, and schedule the deletion. I understand you only wrote this to see how often it was happening we were adding an order when it was not time

src/domain/events/index.ts
```diff
@@ -245,6 +268,7 @@ export function add(
         },
       ])
     );
+    // TODO: why twice?
```
anxolin, line 271: One increment owners and one orders

src/domain/events/index.ts
```diff
@@ -256,19 +280,20 @@ export function add(
...
-          // Delete the conditional order
-          conditionalOrders.delete(conditionalOrder);
-        }
-      }
+function flush(owner: Owner, root: BytesLike, registry: Registry) {
+  if (!registry.ownerOrders.has(owner)) return;
+
+  const conditionalOrders = registry.ownerOrders.get(owner);
+
+  if (conditionalOrders === undefined) return;
```
anxolin, line 288: simpler, good call

src/types/model.ts
```diff
@@ -95,27 +96,32 @@ export interface RegistryBlock {
   hash: string;
 }
 
+type OrdersPerOwner = Map<Owner, Set<ConditionalOrder>>;
+
+const logger = getLogger("Registry");
```
anxolin, line 101: Why registry if this is in `model.ts`?

src/utils/misc.ts
```diff
@@ -31,3 +31,14 @@ export function toConditionalOrderParams({
...
+
+export function getAreConditionalOrderParamsEqual(
+  a: ConditionalOrderParams,
+  b: ConditionalOrderParams
+): boolean {
+  return (
+    a.handler === b.handler &&
+    a.salt === b.salt &&
+    a.staticInput === b.staticInput
+  );
```
anxolin, line 43: Do we care about case sensitivity?

src/domain/events/index.ts
shoom3301: I don't know, I just moved this code from the `_addContract` function in order to not create excessive instances

src/domain/events/index.ts
shoom3301: I didn't change this code, just removed it from the exports and removed `_` suffix

src/domain/events/index.ts
shoom3301: Added

src/domain/events/index.ts
shoom3301: Thanks!
Changed it to:

```
// One increment owners and one orders
    metrics.activeOwnersTotal.labels(network).inc(2);
```

src/utils/misc.ts
shoom3301: I don't think this is a case, but can't 100% be sure.
So, I would better to lowercase them

src/domain/events/index.ts
anxolin: I thought so. In principle are instances that are scoped to that function, so the garbage collector should be able to find (unless we have a leak in the logging library)

If you really want to do this, at least we need to find a better name than just `log`. I might want to add another function `doSomething()` and use that log, and I will see `addContract` in my log messages

src/domain/events/index.ts
anxolin: Yes, but you created a global const with the `composableCow` instance, so no need to pass it as a parameter any more

src/types/model.ts
anxolin: Its a similar comment as above. You took it out of the registry and has still a very generic name. 
I don't think getting loggers should be the memory leak (i hope)

shoom3301: Got you.
Created a separate logger:
```
const loadOwnerOrdersLogger = getLogger("loadOwnerOrders");
```

src/domain/events/index.ts
shoom3301: Got you, reverted the logger back and kept only `composableCow`

anxolin (approved):

PR #165, opened 2024-12-03
fix: fix crash with error in log

anxolin:
# Description
We have a crash in watch-tower staging env after https://github.com/cowprotocol/watch-tower/pull/164

see 
https://aws-es.cow.fi/_dashboards/app/data-explorer/discover/#/view/70561750-b0ee-11ef-8199-15daae69bf9c?_q=(filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'5ed837d0-4a67-11ef-8b11-a98c7c46873d',key:kubernetes.pod_labels.app,negate:!f,params:(query:watch-tower),type:phrase),query:(match_phrase:(kubernetes.pod_labels.app:watch-tower)))),query:(language:kuery,query:''))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15h,to:now))&_a=(discover:(columns:!(log,log_level),isDirty:!f,savedSearch:'70561750-b0ee-11ef-8199-15daae69bf9c',sort:!()),metadata:(indexPattern:'5ed837d0-4a67-11ef-8b11-a98c7c46873d',view:discover))

image

PR #166, opened 2024-12-03
fix: fix circular dependency issue

shoom3301:
# Description

1. We import `import { getLogger } from "../utils";` in `types/model.ts`
2. `export * from "./model";`  is in `types/index.ts`
3. `types/index.ts` is used in `utils/misc.ts`
4. `utils/misc.ts` is exported from `"../utils"`


anxolin (approved):
Works! 

image

Closing then https://github.com/cowprotocol/watch-tower/pull/165 in favour of your PR as you found the cause

PR #168, opened 2024-12-03
fix: compare order params properly

shoom3301:
The hypothesis works, so we can confidently add the `getAreConditionalOrderParamsEqual()` call

image


anxolin (approved):
Cool, but you think this was happening in re-orgs (when we re-apply an order) or somehow it was happening more often?

PR #171, opened 2024-12-07
feat: add sync status and time from last indexed block

anxolin:
# Description
One of the things is super hard to monitor in Watch Tower is the sync status. 

## The problem
Syncing from the origin of times (deployment of composable cow), can take a while.  Currently, Its super hard to see which instances of watch-tower are in sync.

Further more, if there's a crash, or you restart a pod, it will enter in sync mode (to catch up). There could be issues of not catching up fast enough. Its very hard to see that your watch-tower is lagging behind.

# Additions in this PR
## Sync status metric
Now, there's a new prometheus metric for the SYNC status.

```
# HELP watch_tower_sync_status Sync status of watch rower. 1 if in sync, 0 otherwise
# TYPE watch_tower_sync_status gauge
watch_tower_sync_status{chain_id="1"} 0
```

## Last block
Once we know we are SYNCING, its hard to tell how far we are from the tip of the blockchain. 

We have some metric to know the last indexed block, but this is hard to use to get a quick idea if we are catching up. With the block you see a growing number, that then you need to check how far from the tip is in a block explorer.

This PR creates a new metric with the timestamp of the latest block. This way, we can present the distance to NOW, to show how far we are from a full sync. Also, as we see the updates in the dashboard real time, we can see the rate in which this relative time decreases.

```
# HELP watch_tower_block_timestamp Block timestamp of the last processed block
# TYPE watch_tower_block_timestamp gauge
watch_tower_block_timestamp{chain_id="1"} 1722003767
```

### DIFF
--- a/src/services/chain.ts
+++ b/src/services/chain.ts
@@ -78,7 +78,7 @@ export class ChainContext {
   readonly addresses?: string[];
   readonly processEveryNumBlocks: number;
 
-  private sync: ChainSync = ChainSync.SYNCING;
+  private sync: ChainSync;
   static chains: Chains = {};
 
   provider: providers.Provider;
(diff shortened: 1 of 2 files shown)


anxolin:
Please @cowprotocol/frontend , i will need  a review on this, but merging cause I need to test this in DEVELOPMENT asap

anxolin:
These are the additions to the grafana charts

image

You can filter also by chain and env:

image

alfetopito (commented):
👍

PR #172, opened 2024-12-07
feat: add metric for the latest block from the chain (tip)

anxolin:
# Description

Follow up on https://github.com/cowprotocol/watch-tower/pull/171

This PR adds the latest mined block (tip of the chain).

It allow us to build a burn-down chart on how far we are from the chain (as we consume blocks).

This is useful to see that indeed the slope is downwards.
Otherwise, we won't arrive to the tip of the chain! 😄 

# Changes
Include a new metric

```

# HELP watch_tower_block_height_latest Block height of the last block (tip of the chain)
# TYPE watch_tower_block_height_latest gauge
watch_tower_block_height_latest{chain_id="1"} 21351273

```

### DIFF
--- a/src/services/chain.ts
+++ b/src/services/chain.ts
@@ -189,7 +189,11 @@ export class ChainContext {
     let fromBlock = lastProcessedBlock
       ? lastProcessedBlock.number + 1
       : this.deploymentBlock;
+
     let currentBlock = await provider.getBlock("latest");
+    metrics.blockHeightLatest
+      .labels(chainId.toString())
+      .set(currentBlock.number);
 
     let printSyncInfo = true; // Print sync info only once
     let toBlock: "latest" | number = 0;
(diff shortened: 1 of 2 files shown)


anxolin:
@cowprotocol/frontend , merging to test this, but please review anyways

src/services/chain.ts
```diff
@@ -189,7 +189,11 @@ export class ChainContext {
     let fromBlock = lastProcessedBlock
       ? lastProcessedBlock.number + 1
       : this.deploymentBlock;
+
     let currentBlock = await provider.getBlock("latest");
+    metrics.blockHeightLatest
+      .labels(chainId.toString())
+      .set(currentBlock.number);
```
alfetopito, line 196: Nitpick: since this is repeated 3 times, would it be worth it to wrap around a fn? Just so we don't forget to track the metrics if we want to use this elsewhere.

anxolin: sorry, i saw it only after merging. The whole code needs some serious refactoring. Lets clean it up all in the future 🙏

PR #176, opened 2024-12-10
feat: apply filter policy before adding an order

anxolin:
# Description
We are not respecting the filter policy in order to add or not add orders from events.

## Context
I noticed how staging indexed this an order of an owner that shouldn't have. However, in the next block got dropped because we apply the filtering using the policies at the beginning of processing a block.

This explain this increment + decreement:
image

You can also see this in the logs:
image

## Chages
This pr applies the filter policy before adding an order, so we don't add it in the first place. 

## Test
You can create one TWAP using any account and see how now we don't get this increment and decrement in Grafana.

### DIFF
--- a/src/domain/events/index.ts
+++ b/src/domain/events/index.ts
@@ -17,13 +17,10 @@ import {
   Proof,
   Registry,
 } from "../../types";
-import {
-  ConditionalOrder,
-  ConditionalOrderParams,
-  SupportedChainId,
-} from "@cowprotocol/cow-sdk";
+import { ConditionalOrder, ConditionalOrderParams } from "@cowprotocol/cow-sdk";
 
 import { ChainContext } from "../../services/chain";
+import { policy } from "../polling/filtering";
 
 const composableCow = ComposableCoW__factory.createInterface();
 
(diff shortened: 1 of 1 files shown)


src/domain/events/index.ts
```diff
@@ -208,15 +207,33 @@ function addOrder(
...
   const log = getLogger({ name: "addOrder", chainId, blockNumber });
   const { handler, salt, staticInput } = params;
   const { network, ownerOrders } = registry;
 
+  const conditionalOrderParams = toConditionalOrderParams(params);
   const conditionalOrderId = ConditionalOrder.leafToId(params);
+
+  // Apply the filter policy, in case this order should be dropped
+  if (filterPolicy) {
+    const filterResult = filterPolicy.preFilter({
```
anxolin, line 223: key part of this PR. We apply the filtering.

src/domain/events/index.ts
```diff
@@ -208,15 +207,33 @@ function addOrder(
...
+      conditionalOrderId,
+      transaction: tx,
+      owner,
+      conditionalOrderParams,
+    });
+
+    if (filterResult === policy.FilterAction.DROP) {
```
anxolin, line 230: we only handle DROP, because for SKIP we should add the order

src/domain/events/index.ts
```diff
@@ -208,15 +207,33 @@ function addOrder(
   params: ConditionalOrderParams,
   proof: Proof | null,
   composableCow: string,
-  registry: Registry,
-  chainId: SupportedChainId,
-  blockNumber: number
-) {
+  blockNumber: number,
+  context: ChainContext
```
alfetopito, line 211: Please update the docstring as well

alfetopito (approved):

PR #207, opened 2026-07-22
feat: log memory usage and shutdown

shoom3301:
1. Add a `SIGTERM` handler (mirroring the SIGINT one). Next time it will log a clean shutdown on any k8s-initiated restart, so silence will then unambiguously mean SIGKILL/OOM. Right now the two cases are indistinguishable.
2. Periodically log memory usage to see when OOM happens

## Summary by CodeRabbit

* **Monitoring**
  * Added periodic memory usage logging (RSS, heap, and external), reported on a timed cadence with values formatted in megabytes.

* **Bug Fixes**
  * Improved shutdown handling so the memory monitoring stops cleanly on normal completion and on error conditions.
  * Added graceful shutdown support for both SIGINT and SIGTERM, ensuring termination triggers exit after stopping monitoring.

### DIFF
--- a/src/commands/run.ts
+++ b/src/commands/run.ts
@@ -5,7 +5,7 @@ import { DBService, ApiService, ChainContext } from "../services";
 // How often to log the process memory usage. This leaves a memory trail in the
 // logs so a restart can be correlated with memory pressure (e.g. OOM) even when
 // only the application logs are available.
-const MEMORY_LOG_FREQUENCY_SECS = 60;
+const MEMORY_LOG_INTERVAL_MS = 60_000;
 
 /**
  * Run the watch-tower 👀🐮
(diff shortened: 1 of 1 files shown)


src/commands/run.ts
```diff
@@ -2,6 +2,11 @@ import { RunOptions } from "../types";
 import { getLogger } from "../utils";
 import { DBService, ApiService, ChainContext } from "../services";
 
+// How often to log the process memory usage. This leaves a memory trail in the
+// logs so a restart can be correlated with memory pressure (e.g. OOM) even when
+// only the application logs are available.
+const MEMORY_LOG_FREQUENCY_SECS = 60;
```
limitofzero: nitpick: I think it's actually is a period or interval then frequency

limitofzero (approved):

azebuado:
@coderabbitai review

azebuado (approved):

PR #209, opened 2026-07-28
fix: fix crashlooping causes

shoom3301:
## Summary

Fixes a ~20 minute restart loop on `base-watch-tower-develop` (~160 container restarts observed over ~34h). The pod was not OOM-killed — it was killed by the liveness probe:

1. Each block pass took ~40s (~22s posting orders, ~18s persisting a 27.5 MB registry) while only ~40s of chain time elapsed, so the watcher sat permanently ~300s behind the head.
2. Any pass that overran tripped the watchdog, which set `sync = UNKNOWN`.
3. Nothing ever set it back — both `IN_SYNC` assignments lived inside `warmUp()` — so `/health` returned 503 forever and the probe killed the pod, even though the watcher recovered within ~20-30s.

### 1. Registry pruning — the main performance fix

`ConditionalOrder.orders` is an append-only `Map<OrderUid, OrderStatus>` used purely to dedup posts. Nothing ever removed entries and `FILLED` was never set, so it had grown to **27.5 MB for 65 orders** (~220k stale entries), re-serialised and rewritten on every block.

`pruneExpiredOrders()` decodes `validTo` from the trailing 4 bytes of the GPv2 order uid (`digest(32) || owner(20) || validTo(4)`) and drops entries that can never be posted again. Uids are validated in full against `^0x[0-9a-fA-F]{112}$` before parsing — `parseInt` stops at the first invalid character rather than returning `NaN`, so length-only validation would silently prune corrupt entries.

Pruning runs **once per block** at the start of `checkForAndPlaceOrder`, not inside `write()` — `write()` is also called every `CHUNK_SIZE` (50) orders, and a whole-registry scan there would repeat dozens of times per block. Running it before the loop also means the first post-deploy chunked write no longer serialises the full expired registry.

This removes the ~18s write and the memory sawtooth (`heapUsed` 70 → 183 MB per cycle was the 27.5 MB stringify).

### 2. Watchdog recovery

`watchdogSyncState()` returns `IN_SYNC` when `timeElapsed < watchdogTimeout`, and the watchdog loop now assigns it, logging `Chain watcher recovered`. A transient stall no longer permanently fails the liveness probe. Non-Kubernetes exit behaviour is unchanged.

**`WATCHDOG_TIMEOUT_DEFAULT_SECS` raised 30 → 60.** This doubles how long a stalled chain goes undetected for deployments using the default, and is required so `ORDER_BOOK_API_TIMEOUT_MS` can sit below it.

### 3. Retry policy for API rejections

`escalatingRetryDelay()`: attempts 1–2 retry next block, then 1 min → 10 min → 1 hr. Tracked per order via `consecutiveApiFailures`, **scoped by `lastApiError`** so a run of one error type doesn't push an unrelated error straight to the top tier. Reset on acceptance, including `DUPLICATED_ORDER`.

**`InvalidEip1271Signature` is deliberately exempt.** It is the expected response while a presign transaction is unexecuted, which for a Safe can legitimately last days. Those orders keep polling **every block** so they are placed the moment the presign lands. They are only dropped once the conditional order is older than `PRESIGN_MAX_AGE_SECS` (5 days), at which point the presign is assumed never to be coming. `createdAtEpoch` is backfilled on first poll, so orders persisted before this field existed start their clock at deploy rather than being exempt forever.

Timeouts route through the same progression rather than returning `UNEXPECTED_ERROR`.

### 4. Timeouts and retries for external calls

New `withTimeout()` and `withRetry()` helpers.

| Path | Bound | Rationale |
|---|---|---|
| `sendOrder` | 40s | ~12x observed p99 (3.4s); must stay under the 60s watchdog — enforced by a test, not a comment |
| `getLogs` (block path) | 30s | Serialised behind the watchdog, must fail fast |
| warm-up RPC | 180s + 5 retries w/ backoff | Pages over `pageSize` blocks with no watchdog running; a cold-start backfill is legitimately slow |

The warm-up bound is deliberately much looser than the block path. Retrying does not help an operation that is *consistently* slower than its deadline — every attempt hits the same wall, and the failure propagates to `process.exit(1)`, taking down every other chain. `withTimeout` clears its timer on both paths (verified by removing `clearTimeout` and confirming the test caught the leaked handle).

## Notes

**No migration required.** `createdAtEpoch`, `consecutiveApiFailures` and `lastApiError` are all optional; pruning doesn't change the storage format. Registry version stays at v2.

**First pass after deploy is expensive**: one ~220k-entry scan plus a 27.5 MB serialise. Every block after that is cheap.

**Known limitations, not addressed here:**
- Per-call timeouts bound a single call, not a whole block pass. A pass posting many slow orders can still exceed the watchdog.
- Per-order backoff can't distinguish "this order is broken" from "the API is down". A circuit breaker keyed on the API would recover faster from an outage.
- `withTimeout` abandons rather than cancels; the underlying request stays in flight until the socket closes.
- `InvalidAppData` keeps its existing fixed 60s backoff via `API_ERRORS_BACKOFF`.

## Tests

100 tests (up from 60), covering `escalatingRetryDelay`, `pruneExpiredOrders` (including malformed and non-hex uids), `watchdogSyncState`, `withTimeout`, `withRetry`, presign exemption and cut-off, failure-counter scoping, the chunked-write ordering, and the constant invariants that keep the timeouts from drifting apart.

## Summary by CodeRabbit

- **Reliability**
  - Added time limits for order submission and blockchain RPC operations, preventing stalled processing.
  - Added automatic retries with increasing delays for temporary failures.
  - Improved watchdog recovery when block processing resumes.

- **Order Management**
  - Expired discrete orders are now removed automatically.
  - Order age is tracked consistently.
  - Repeated submission failures are tracked and handled with progressively later retries.

- **Error Handling**
  - Timeout failures are reported distinctly from unexpected errors, improving retry behavior and diagnostics.

### DIFF
--- a/src/domain/polling/index.ts
+++ b/src/domain/polling/index.ts
@@ -172,6 +172,19 @@ export async function checkForAndPlaceOrder(
   const log = getLogger(loggerParams);
   log.debug(`The registry has ${numOwners} owners and ${numOrders} orders`);
 
+  // Evict discrete orders that can no longer be placed. `orders` is otherwise
+  // append-only and dominates the persisted payload, which every write pays
+  // for twice - once to `JSON.stringify` and once to write to LevelDB.
+  //
+  // This is a whole-registry scan, so it runs once per block rather than
+  // inside `write()`. It must happen before the loop below, because that also
+  // writes every CHUNK_SIZE orders - pruning afterwards would leave the first
+  // post-deploy chunks serialising the full expired registry.
+  const pruned = registry.prune(blockTimestamp);
+  if (pruned > 0) {
+    log.debug(`Pruned ${pruned} expired discrete orders`);
+  }
+
   for (const [owner, conditionalOrders] of ownerOrders.entries()) {
     ownerCounter++;
     const log = getLogger({
@@ -325,17 +338,6 @@ export async function checkForAndPlaceOrder(
       ownerOrders.delete(owner);
       metrics.activeOwnersTotal.labels(chainId.toString()).dec();
     }
-  }
-
-  // Evict discrete orders that can no longer be placed. `orders` is otherwise
-  // append-only and dominates the persisted payload, which every write pays
-  // for twice - once to `JSON.stringify` and once to write to LevelDB.
-  //
-  // This is a whole-registry scan, so it runs once per block here rather than
-  // inside `write()`, which is also called on every CHUNK_SIZE orders.
-  const pruned = registry.prune(blockTimestamp);
-  if (pruned > 0) {
-    log.debug(`Pruned ${pruned} expired discrete orders`);
   }
 
   // save the registry - don't catch errors here, as it's now a docker container
--- a/src/domain/polling/index.ts
+++ b/src/domain/polling/index.ts
@@ -130,10 +130,16 @@ const CHUNK_SIZE = 50; // How many orders to process before saving
 
 /**
  * Upper bound on a single `sendOrder` call, including the SDK's internal
- * retries. Comfortably above the observed p99 (~3s) but far below the
- * watchdog timeout, so a wedged socket can't stall block processing.
+ * retries.
+ *
+ * Must stay below `WATCHDOG_TIMEOUT_DEFAULT_SECS` so one wedged socket cannot
+ * on its own push the chain watcher past the watchdog deadline. Still ~6x the
+ * observed p99 (~3.4s), leaving room for the SDK to retry transport failures.
+ *
+ * Note this bounds a *single* call, not a whole block pass - a pass posting
+ * many orders can still exceed the watchdog if each one is slow.
  */
-const ORDER_BOOK_API_TIMEOUT_MS = 60_000;
+export const ORDER_BOOK_API_TIMEOUT_MS = 40_000;
 
 /**
  * Watch for new blocks and check for orders to place
@@ -319,6 +325,17 @@ export async function checkForAndPlaceOrder(
       ownerOrders.delete(owner);
       metrics.activeOwnersTotal.labels(chainId.toString()).dec();
     }
+  }
+
+  // Evict discrete orders that can no longer be placed. `orders` is otherwise
+  // append-only and dominates the persisted payload, which every write pays
+  // for twice - once to `JSON.stringify` and once to write to LevelDB.
+  //
+  // This is a whole-registry scan, so it runs once per block here rather than
+  // inside `write()`, which is also called on every CHUNK_SIZE orders.
+  const pruned = registry.prune(blockTimestamp);
+  if (pruned > 0) {
+    log.debug(`Pruned ${pruned} expired discrete orders`);
   }
 
   // save the registry - don't catch errors here, as it's now a docker container
@@ -567,7 +584,7 @@ export const _printUnfilledOrders = (orders: Map<BytesLike, OrderStatus>) => {
  * @param order to be placed on the cow protocol api
  * @param apiUrl rest api url
  */
-async function postDiscreteOrder(params: {
+export async function postDiscreteOrder(params: {
   conditionalOrder: ConditionalOrder;
   orderUid: string;
   order: any;
@@ -677,8 +694,9 @@ async function postDiscreteOrder(params: {
       // http.ClientRequest in node.js
       reasonError += `Unresponsive API: ${error.request}`;
     } else if (error.message) {
-      // Something happened in setting up the request that triggered an Error
-      reasonError += `. Internal Error: ${error.request}`;
+      // Something happened in setting up the request that triggered an Error,
+      // including a `TimeoutError` from the `withTimeout` wrapper.
+      reasonError += `. Internal Error: ${error.message}`;
     } else {
       reasonError += `. Unhandled Error: ${error.message}`;
     }
--- a/src/domain/polling/index.ts
+++ b/src/domain/polling/index.ts
@@ -32,6 +32,7 @@ import {
   getLogger,
   handleOnChainCustomError,
   metrics,
+  withTimeout,
 } from "../../utils";
 import { badOrder, policy } from "./filtering";
 import { pollConditionalOrder } from "./poll";
@@ -72,6 +73,37 @@ const API_ERRORS_BACKOFF: BackOffApiErrorsDelays = {
   [ApiErrors.INVALID_APP_DATA]: ONE_MIN, // Give the user some time to upload the correct appData
 };
 
+/** How many times to retry on the very next block before backing off */
+const TRY_NEXT_BLOCK_MAX_ATTEMPTS = 2;
+/** Backoff applied once `TRY_NEXT_BLOCK_MAX_ATTEMPTS` is exhausted */
+const ESCALATING_BACKOFF_DELAYS = [ONE_MIN, TEN_MINS, ONE_HOUR];
+
+/**
+ * Escalating backoff for API errors that are *usually* transient but, when
+ * they aren't, would otherwise be retried on every single block forever.
+ *
+ * @param consecutiveFailures how many times in a row this order has failed,
+ * including the current failure
+ * @returns seconds to wait before the next attempt, or `undefined` to retry on
+ * the next block
+ */
+export function escalatingRetryDelay(
+  consecutiveFailures: number
+): number | undefined {
+  if (consecutiveFailures <= TRY_NEXT_BLOCK_MAX_ATTEMPTS) {
+    return undefined;
+  }
+
+  const delay =
+    ESCALATING_BACKOFF_DELAYS[
+      consecutiveFailures - TRY_NEXT_BLOCK_MAX_ATTEMPTS - 1
+    ];
+
+  return (
+    delay ?? ESCALATING_BACKOFF_DELAYS[ESCALATING_BACKOFF_DELAYS.length - 1]
+  );
+}
+
 const API_ERRORS_DROP: DropApiErrorsArray = [
   ApiErrors.SELL_AMOUNT_OVERFLOW, // Implies a `feeAmount` has been set and `sellAmount` + `feeAmount` > `type(uint256).max`
   ApiErrors.TRANSFER_SIMULATION_FAILED, // Sell token can't be transferred, drop it
@@ -95,6 +127,13 @@ const API_ERRORS_DROP: DropApiErrorsArray = [
 // ApiErrors.AppDataHashMismatch - we never submit full appData
 
 const CHUNK_SIZE = 50; // How many orders to process before saving
+
+/**
+ * Upper bound on a single `sendOrder` call, including the SDK's internal
+ * retries. Comfortably above the observed p99 (~3s) but far below the
+ * watchdog timeout, so a wedged socket can't stall block processing.
+ */
+const ORDER_BOOK_API_TIMEOUT_MS = 60_000;
 
 /**
  * Watch for new blocks and check for orders to place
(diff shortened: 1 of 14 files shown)


src/domain/polling/index.ts
```diff
@@ -72,6 +73,37 @@ const API_ERRORS_BACKOFF: BackOffApiErrorsDelays = {
...
+export function escalatingRetryDelay(
+  consecutiveFailures: number
+): number | undefined {
+  if (consecutiveFailures <= TRY_NEXT_BLOCK_MAX_ATTEMPTS) {
+    return undefined;
+  }
+
+  const delay =
+    ESCALATING_BACKOFF_DELAYS[
+      consecutiveFailures - TRY_NEXT_BLOCK_MAX_ATTEMPTS - 1
```
azebuado, line 100: Isn't this a bit brittle? Can't `consecutiveFailures` be over 5, going out of bounds?

src/domain/polling/index.ts
```diff
@@ -587,23 +626,41 @@ async function postDiscreteOrder(params: {
     );
     log.debug(`Post order ${orderUid} details`, postOrder);
     if (!dryRun) {
-      const orderUid = await orderBookApi.sendOrder(postOrder);
+      const orderUid = await withTimeout(
+        orderBookApi.sendOrder(postOrder),
+        ORDER_BOOK_API_TIMEOUT_MS,
+        `sendOrder (ID=${conditionalOrder.id})`
+      );
```
azebuado, line 690: I remember seeing this before but I don't recall the reason.
Are you able to tell why those requests can take longer than 1min?

src/services/chain.ts
```diff
@@ -614,11 +631,17 @@ async function pollContractForEvents(
...
   const topic = ethers.utils.id(eventName);
 
-  const logs = await provider.getLogs({
-    fromBlock,
-    toBlock,
-    topics: [topic],
-  });
+  // Bound the RPC call - block processing is serialised, so a hung `getLogs`
+  // stalls every block queued behind it.
+  const logs = await withTimeout(
```
azebuado: Is this retried if a call fails? Does it matter?

src/types/model.ts
```diff
@@ -91,13 +100,64 @@ export type ConditionalOrder = {
...
+/** `"0x"` + 32 byte order digest + 20 byte owner + 4 byte `validTo` */
+const ORDER_UID_HEX_LENGTH = 2 + 2 * (32 + 20 + 4);
+const VALID_TO_HEX_LENGTH = 2 * 4;
+
+/**
+ * Read the `validTo` encoded in the trailing 4 bytes of a GPv2 order uid.
+ * Returns `undefined` for anything that isn't a well formed uid, so unknown
+ * entries are retained rather than silently discarded.
+ */
+function getOrderUidValidTo(orderUid: OrderUid): number | undefined {
```
azebuado: Given that we use that to check whether the order should be placed or not, I'm surprise a fn like this didn't exist before.

azebuado (commented):
My own comments

azebuado:
<details>
  <summary>⚠️ <strong>AI Review</strong> (Codex GPT-5, worked 6m): malformed UIDs can be deleted</summary>

  Finding: [NON-BLOCKING] Preserve full-length malformed order UIDs

  - Location: src/types/model.ts:getOrderUidValidTo (line 141)
  - parseInt(..., 16) accepts a valid prefix. A 114-character UID ending in 0000000g is parsed as 0, then pruned as expired.
  - This contradicts the stated contract to retain malformed UIDs and can discard unknown persisted registry state.

  Suggested fix

  - Validate the complete UID as hex before parsing (for example, ^0x[0-9a-fA-F]{112}$).
  - Add a regression test for a full-length UID with a non-hex character.
  <details>
  <summary>Review scope and related context</summary>

  - Existing comments still cover the request deadlines exceeding—or equaling—the default 30s watchdog, timeout errors bypassing the new backoff, and unbounded warm-up RPCs. Those should
    be resolved or explicitly closed before merge.

  - Rechecked the human concern about failure counts above five: the hourly fallback handles it correctly.
  - Automated validation: yarn test passed (80 tests); yarn lint completed with 0 errors.
  - Security: DeepSec completed a direct scan of the five changed production TypeScript files with 0 findings; code inspection found no verified PR-introduced security vulnerability.
  </details>
  <details>
  <summary>🤖 Prompt for AI agents</summary>

  Fix only the UID-validation bug in src/types/model.ts.

  Before parsing validTo, require a complete GPv2 UID: 0x followed by 112 hex characters. Preserve malformed entries unchanged. Add a test for a full-length UID with a non-hex final
  character and verify it is retained.
  </details>

  Generated using the pr-review and security-review skills from the CoW Protocol skills repo (https://github.com/cowprotocol/skills).
  </details>

src/services/chain.ts
azebuado: Additional notes by Claude which I think it's related.

1. getLogs timeout can silently drop ConditionalOrderCreated events (High) — src/services/chain.ts:352-362, 343-395
pollContractForEvents(blockNumber, blockNumber, this) is called with a single fixed block range inside subscribeToNewBlocks's per-block handler. If it throws (now including the new 30s TimeoutError), the outer try/catch (chain.ts:387-394) just logs and moves on — processBlockAndPersist/persistLastProcessedBlock is never called for that block, so registry.lastProcessedBlock never advances past it. Nothing else re-fetches that specific range: getEventPollingRange catch-up only runs in warmUp() at startup. Meanwhile the watchdog's lastBlockReceived is updated unconditionally before the getLogs call (chain.ts:378), so /health stays green throughout.

Net effect: before this PR, a genuinely wedged getLogs call hung the process, eventually triggering a k8s restart, which self-healed via warmUp()'s range replay. After this PR, the same wedge times out after 30s, is swallowed, and the block's events are permanently lost with no restart and no health-check signal — a new conditional order created in that block would simply never be watched. Worth adding a targeted re-fetch/backfill (or at least tracking a gap and forcing a warmUp-style catch-up) rather than relying on process death as the (accidental) recovery mechanism.

src/types/model.ts
```diff
@@ -205,9 +265,35 @@ export class Registry {
...
+
+    return pruned;
+  }
+
   public async write(): Promise<void> {
+    // Evict discrete orders that can no longer be placed before serialising.
+    // `orders` is otherwise append-only and dominates the persisted payload -
+    // left unbounded it grows into tens of megabytes, and every block pays for
+    // it twice, once to `JSON.stringify` and once to write to LevelDB.
+    const nowEpoch = this.lastProcessedBlock?.timestamp;
```
azebuado: Additional review notes by Claude.

2. Registry.prune() runs on every chunked write, not once per block (Medium) — src/types/model.ts:648-660, src/domain/polling/index.ts:154-164
registry.write() is called every CHUNK_SIZE (50) processed orders and once at the end of checkForAndPlaceOrder. Pruning now happens inside write() itself and scans every owner's every conditional order's orders map each time. With a large registry (the PR's own numbers: ~220k entries), a block with thousands of active conditional orders will call write() dozens/hundreds of times, each doing a full-registry scan — turning what should be O(n) work per block into something closer to O(n²/CHUNK_SIZE). This risks reintroducing a CPU-bound version of the exact perf problem being fixed, especially on the first pass after deploy where the scan is largest. Consider pruning once per checkForAndPlaceOrder call (or gating so only the final write() in a block prunes) instead of on every chunk.

src/domain/polling/index.ts
shoom3301: They can, because of that there is a fallback below:
```
delay ?? ESCALATING_BACKOFF_DELAYS[ESCALATING_BACKOFF_DELAYS.length - 1]
```

src/domain/polling/index.ts
shoom3301: There might various of reasons:
 - network conditions
 - OrderBookApi internal errors

shoom3301: p.s. changed the timeout to 40s

src/services/chain.ts
shoom3301: If it fails, then `lastBlockReceived` won't be bumped, and the next iteration will try again from the last successful block

shoom3301: But your concern has a basement.
There are two places where we call `pollContractForEvents`: `subscribeToNewBlocks` and `warmUp`.
For the warmUp we  should not add the timeout. Thanks! Fixed

src/utils/timeout.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(operation: string, timeoutMs: number) {
+    super(`${operation} timed out after ${timeoutMs}ms`);
+    this.name = "TimeoutError";
+  }
+}
+
+/**
+ * Bound how long an operation may take.
+ *
+ * Neither the orderbook SDK nor the ethers providers apply a request timeout,
```
limitofzero: the comment is not actually true, I've asked claude to check if there is a timeout in ether/sdk, ether has - 120s. 

unfortunately not ether neither sdk don't accept AbortController :( socket connection is still alive after the timeout and it also increase a memory consumption

shoom3301:
@coderabbitai review

src/services/chain.ts
shoom3301: Added retries

shoom3301:
@azebuado @limitofzero thank you! Addressed all AI and your comments.

Summary:
1. Improved `getBlock()` calls in `warmUp` with `withRetry` and `withTimeout`. Warm-up RPC calls are retried rather than crashing the whole run.
2. Changed ORDER_BOOK_API_TIMEOUT_MS=40s, RPC_TIMEOUT_MS=30s, and WATCHDOG_TIMEOUT_DEFAULT_SECS=60s. I checked prod configs, and `watchdogTimeout` is actually always higher than 120s.
3. Enhanced validation in `getOrderUidValidTo`
4. Fixed `TimeoutError` logging

src/utils/timeout.ts
shoom3301: Fixed the comment, thanks

src/types/model.ts
shoom3301: Thanks! Removed prune() call from write()

shoom3301: Removed the prune from `Registry.write()`, added a single `registry.prune(blockTimestamp)` in `checkForAndPlaceOrder` immediately before the final write. write() is back to being purely a serialize-and-persist, which is what its callers assume.

src/domain/polling/index.ts
azebuado: Got that, just looks a bit weird.

src/types/model.ts
azebuado: <details>
  <summary>⚠️ <strong>AI Review</strong> (Codex GPT-5, worked 5m): follow-up still reproducible</summary>

  Finding: [BLOCKING] Prune before the first chunked write

  - Location: src/domain/polling/index.ts:198 and :330
  - This one is important: registry.write() runs after every chunk before registry.prune(blockTimestamp) runs at the end.
  - With more than 51 processed orders, the first post-deploy pass still serializes and writes the full expired registry once per chunk. That can repeat the large write/memory spike this
    PR is meant to remove.

  Suggested next step

  - Prune once near the start of checkForAndPlaceOrder, before any chunk can call registry.write().
  - Add coverage with more than CHUNK_SIZE processed orders, asserting the first persisted registry excludes expired UIDs.
  <details>
  <summary>🤖 Prompt for AI agents</summary>

  Move the once-per-block registry.prune(blockTimestamp) call so it happens before the processing loop can issue any chunked registry.write().

  Keep pruning to one whole-registry scan per block. Add a regression test with more than 50 processed conditional orders and expired discrete UIDs, proving the first chunked write
  serializes only live UIDs.
  </details>

  Generated using the pr-review skill from the CoW Protocol skills repo (https://github.com/cowprotocol/skills).
  </details>

src/domain/polling/index.ts
```diff
@@ -587,23 +626,41 @@ async function postDiscreteOrder(params: {
...
-        metricLabels
+        metricLabels,
+        consecutiveFailures
       );
+
+      // A duplicated order means it is already in the orderbook - treat that
+      // as acceptance so we don't back off a perfectly healthy order.
+      if (handleErrorResult.result === PollResultCode.SUCCESS) {
+        conditionalOrder.consecutiveApiFailures = 0;
+      }
```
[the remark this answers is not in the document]
azebuado: <details>
  <summary>⚠️ <strong>AI Review</strong> (Codex GPT-5, worked 5m): follow-up still reproducible</summary>

  Rechecked

  - Location: src/domain/polling/index.ts:657
  - TimeoutError still has neither response nor request, so it reaches the error.message branch and returns UNEXPECTED_ERROR.
  - consecutiveApiFailures is only incremented in the error.response branch, so a timed-out submission receives no escalating backoff.

  Result: Still unresolved. The new test improves the error message but leaves the retry behavior unchanged.

  Suggested next step

  - Handle TimeoutError explicitly and route it through the failure counter plus TRY_NEXT_BLOCK / escalating TRY_AT_EPOCH result.
  - Add a timeout-specific test, rather than only a generic rejected request test.
  <details>
  <summary>🤖 Prompt for AI agents</summary>

  In postDiscreteOrder, detect TimeoutError before the response/request branches.

  Increment consecutiveApiFailures and return the same retry/backoff progression used for transient order-book failures. Add a test using a promise that times out and assert it does not
  return UNEXPECTED_ERROR.
  </details>

  Generated using the pr-review skill from the CoW Protocol skills repo (https://github.com/cowprotocol/skills).
  </details>

src/types/model.ts
shoom3301: Thanks!
Moved `registry.prune()` from just before the final write to immediately after the registry-size debug line, ahead of the owner loop

src/domain/polling/index.ts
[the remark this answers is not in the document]
shoom3301: Thanks!
`TimeoutError` is now detected first in `postDiscreteOrder` catch. It increments `consecutiveApiFailures` and returns the same progression as transient order-book failures — TRY_NEXT_BLOCK for the first two, then TRY_AT_EPOCH.

https://github.com/cowprotocol/watch-tower/pull/209/commits/b0a58dbfce50484b3522b8c7e1de12c79e39fc58

shoom3301:
@coderabbitai review

src/services/chain.ts
```diff
@@ -188,6 +212,24 @@ export class ChainContext {
   public async warmUp(oneShot?: boolean) {
     const { provider, chainId, processEveryNumBlocks } = this;
     const log = getLogger({ name: "warmUp", chainId });
+
+    // Warm-up runs before the watchdog starts, and any error escaping here
+    // reaches `run()` and exits the process - taking down every other chain
+    // with it. So bound each RPC call and retry transient failures with
+    // backoff, isolating a flaky RPC to the chain it belongs to.
```
azebuado, line 219: To be clear, we run one instance per chain, so there's no risk of taking other chains down.

azebuado (approved):

